PR #17, opened 2020-02-20
NOISSUE: log marshalling for errors

cyraxred:
Use structs for errors same as for logger, plus unified handling of panics, stack traces etc


ledger-core/v2/log/adapters/bilog/encoder.go
```diff
@@ -6,6 +6,8 @@
 package bilog
 
 import (
+	"github.com/insolar/assured-ledger/ledger-core/v2/log/logoutput"
```
sexyhamster: something wrong with imports order

ledger-core/v2/vanilla/reflectkit/func_code.go
```diff
@@ -50,3 +50,23 @@ func unwrapIface(v interface{}) (word unsafe.Pointer, kind uint8) {
...
 	return iface.word, iface.typ.kind
 }
+
+// A faster version (25-35%) to check for interface nil
+func IsNil(v interface{}) bool {
+	if v == nil {
+		return true
+	}
+	ptr, kind := unwrapIface(v)
+	return ptr == nil && kind&kindDirectIface != 0
```
sexyhamster: This is really bad code because it relies on assumptions that are not part of spec. It could break after random version change (even minor). Do we really need it?

cyraxred: 1) such change won't be occasional, like with major version that will be carefully evaluated anyway
2) there are tests to ensure behavior on possible cases
3) there is an available fallback option - so even if the code will stop working, it can be replaced without impact to users

ledger-core/v2/log/adapters/bilog/encoder.go
cyraxred: Fixed. GoLand losses settings sometimes :(

cyraxred: oops, this fix will be in a different PR

ledger-core/v2/log/logfmt/object.go
```diff
@@ -17,7 +18,7 @@ type LogObject interface {
 }
 
 type LogObjectMarshaller interface {
-	MarshalLogObject(LogObjectWriter, LogObjectMetricCollector) string
+	MarshalLogObject(LogObjectWriter, LogObjectMetricCollector) (msg string, defMsg bool)
```
sexyhamster: defMsg is not very clear naming. I'm not very clear where do you need the fact defMsg true/false

Maybe API like Msg{Text: string, TxtDefined: bool} would be more clear?

cyraxred: Creating a struct is needless here as these 2 values are only returned, but are never passed together anywhere else.

defMsg means that the msg is generated by default (not provided by user).
A default msg is generated for logger / user convenience and contains struct type name.
But this is not needed for error marshalling, so the bool flag is checked inside the error marshaller to ignore msg when it is default.

This can be changed into an input parameter, but internally the message will be extracted anyway (to reuse same object marshaller for log and for errors). So it will be misleading as an input parameter also.

PR #30, opened 2020-03-17
PLAT-132:  udp pulsar

sexyhamster:
switched pulsar distributor to udp packet.
Moved to proper pulse serialization and now packet could be processed inside consensus.
Packet is rather small, but signature verification is missing for now.


ledger-core/v2/network/consensus/serialization/globula_protocol_test.go
```diff
@@ -253,172 +216,173 @@ func TestGlobulaConsensusPacketBody_Phases(t *testing.T) {
...
-
-			p2 := Packet{}
-
-			_, err = p2.DeserializeFrom(context.Background(), buf)
-			require.NoError(t, err)
-		})
-	}
-}
+//
+// func TestGlobulaConsensusPacketBody_Phases_Flag0(t *testing.T) {
```
sexyhamster: @cyraxred here I need your help, I'm not quite sure what actually is going on here.

old NewPulsePacketWithTrace is protobuf packet, as far as I understand, how and why it goes to .setData. 

It looks like theese tests are no longer valid at all and could be deleted

sexyhamster: @AndreyBronin maybe you know something about it?

sexyhamster: ok, fixed the tests with proper pulsar_packet

ledger-core/v2/network/pulsenetwork/distributor.go
```diff
@@ -67,13 +70,22 @@ func NewDistributor(conf configuration.PulseDistributor) (insolar.PulseDistribut
 }
 
 func (d *distributor) Init(ctx context.Context) error {
-	handler := hostnetwork.NewStreamHandler(func(context.Context, *packet.ReceivedPacket) {}, d.responseHandler)
-
 	var err error
-	d.transport, err = d.Factory.CreateStreamTransport(handler)
+	d.transport, err = d.Factory.CreateDatagramTransport(nil)
```
AndreyBronin: You will get nil pointer dereference, if udp transport will receive a packet. I think you should create empty handler

sexyhamster: hm, only if someone will try to use public address for transport service. Which supposed to be empty anyway, because pulsar should not handle inbound connections (until pulsar network happens)

sexyhamster: I'll create panicHandler with more verbose text

ledger-core/v2/network/consensus/adapters/tests/pulse_test.go
```diff
@@ -56,19 +74,26 @@ func (p *Pulsar) Pulse(ctx context.Context, attempts int) {
...
+	pp := serialization.BuildPulsarPacket(ctx, data)
 
 	go func() {
 		for i := 0; i < attempts; i++ {
-			handler := p.pulseHandlers[rand.Intn(len(p.pulseHandlers))]
-			go handler.HandlePulse(ctx, pu, rp)
+			nodeID := rand.Intn(len(p.addresses))
+			address := p.addresses[nodeID]
+			transport := p.transports[nodeID]
+			go func() {
```
afiskon: This is a potential resource leak. I suggest to consider limiting the total number of goroutines at least somehow, e.g. use an atomic variable that limits the total amount of goroutines to, say, 1000. A metric would be nice too, although this might be not a priority for now.

sexyhamster: it's a test dummy pulsar, It shoots `attempts` goroutines per tick in tests. There are no infinite loops/infinite selects in goroutine so all of them will stop eventually. I don't see resource leak

afiskon: OK, I missed that it's in tests/ directory, sri.

ledger-core/v2/network/pulsenetwork/distributor.go
```diff
@@ -52,6 +52,12 @@ type distributor struct {
 	responseHandler future.PacketHandler
 }
 
+type panicHandler struct{}
```
ruz: `panicHandler` - it's something that handles panics. Use more descriptive name: `panicingHandler`, `handlerThatPanics` or name of your choice that suites.

ledger-core/v2/network/pulsenetwork/distributor.go
```diff
@@ -67,13 +76,22 @@ func NewDistributor(conf configuration.PulseDistributor) (insolar.PulseDistribut
 }
 
 func (d *distributor) Init(ctx context.Context) error {
-	handler := hostnetwork.NewStreamHandler(func(context.Context, *packet.ReceivedPacket) {}, d.responseHandler)
-
 	var err error
-	d.transport, err = d.Factory.CreateStreamTransport(handler)
+	d.transport, err = d.Factory.CreateDatagramTransport(handlerThatPanics{})
```
ruz: I believe this is wrong according to principle of "defensive external APIs". This code shouldn't panic, but log unexpected incoming datagrams.

ledger-core/v2/network/pulsenetwork/distributor.go
```diff
@@ -48,9 +52,14 @@ type distributor struct {
 	responseHandler future.PacketHandler
 }
 
+type handlerThatPanics struct{}
+
+func (ph handlerThatPanics) HandleDatagram(ctx context.Context, address string, buf []byte) {
+	panic("unexpected inbound packet")
```
AndreyBronin: throw.Impossible()

PR #32, opened 2020-03-19
NOISSUE: fix launchnet start

sexyhamster:
Functests still do not work, but at least network starts

### DIFF
--- a/ledger-core/v2/application/testutils/launchnet/launchnet.go
+++ b/ledger-core/v2/application/testutils/launchnet/launchnet.go
@@ -64,6 +64,21 @@ var Enterprise [application.GenesisAmountEnterpriseMembers]*User
 var Foundation [application.GenesisAmountFoundationMembers]*User
 var Funds [application.GenesisAmountFundsMembers]*User
 
+var projectRoot string
+var rootOnce sync.Once
+
+// rootPath returns project root folder
+func rootPath() string {
+	rootOnce.Do(func() {
+		path, err := exec.Command("git", "rev-parse", "--show-toplevel").Output()
+		if err != nil {
+			panic("failed to get project root")
+		}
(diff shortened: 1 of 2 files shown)


ledger-core/v2/application/testutils/launchnet/launchnet.go
```diff
@@ -127,7 +127,7 @@ func launchnetPath(a ...string) (string, error) {
 	cwdList := strings.Split(cwd, "/")
 	var count int
 	for i := len(cwdList); i >= 0; i-- {
-		if cwdList[i-1] == "insolar" && cwdList[i-2] == "insolar" {
+		if cwdList[i-1] == "v2" && cwdList[i-2] == "ledger-core" {
```
afiskon: this line can access cwdList[-1] and cwdList[-2]. I believe for condition should be chaned.

ledger-core/v2/application/testutils/launchnet/launchnet.go
```diff
@@ -415,7 +415,7 @@ func startNet() error {
 		_ = os.Chdir(cwd)
 	}()
 
-	for cwd[len(cwd)-15:] != "insolar/assured-ledger/ledger-core/v2/" {
+	for cwd[len(cwd)-37:] != "insolar/assured-ledger/ledger-core/v2" {
```
afiskon: if len(cwd) < 37 this code will panic.

ledger-core/v2/application/testutils/launchnet/launchnet.go
sexyhamster: yep

ledger-core/v2/application/testutils/launchnet/launchnet.go
sexyhamster: yep

PR #35, opened 2020-03-23
PLAT-133 make new binary insolard_headless

aalekseev-ru:
make new binary insolard_headless that runs nodes without role logic. NB - you still need at least 1 heavy material role

I will add support for launchnet.sh in next task


ledger-core/v2/cmd/insolard_headless/main.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"github.com/spf13/cobra"
+	jww "github.com/spf13/jwalterweatherman"
+
+	"github.com/insolar/assured-ledger/ledger-core/v2/log/global"
+	"github.com/insolar/assured-ledger/ledger-core/v2/server"
+	"github.com/insolar/assured-ledger/ledger-core/v2/version"
+)
+
+func main() {
```
AndreyBronin: Headless network is not a separate binary, it's a mode of insolard binary. See AppController doc

aalekseev-ru: my task is named as "headless network in SEPARATE binary", so I separated them. If it needed to be only a mode I can change this behavior

ledger-core/v2/network/gateway/base.go
```diff
@@ -200,6 +198,8 @@ func (g *Base) createOriginCandidate() error {
 
 func (g *Base) StartConsensus(ctx context.Context) error {
 
+	// one of the nodes has to be in consensus.ReadyNetwork state,
+	// all other nodes has to be in consensus.Joiner
```
ruz: this comment should be inside `if`, I've spend 5 minutes re-reading the comment to understand meaning until I realized that it describes code inside `if`

ledger-core/v2/server/internal/headless/components.go
```diff
@@ -0,0 +1,187 @@
...
+		cryptographyService,
+		keyProcessor,
+	}...)
+
+	// TODO debug
+	// for _, componentMeta := range components {
+	// 	fmt.Print(componentMeta)
+	// 	reflect.ValueOf(componentMeta).Elem()
+	// }
+	cm.Inject(components...)
```
ruz: delete this commented out code

PR #37, opened 2020-04-07
PLAT-194: API for func tests without implementation

ivanshibitov:
**- What I did**
Add test wallet api for to all roles

**- How I did it**

**- How to verify it**


ledger-core/v2/configuration/dummyapi.go
```diff
@@ -0,0 +1,15 @@
+// Copyright 2020 Insolar Network Ltd.
+// All rights reserved.
+// This material is licensed under the Insolar License version 1.0,
+// available at https://github.com/insolar/assured-ledger/blob/master/LICENSE.md.
+
+package configuration
+
+// APIRunner holds configuration for api
+type TestWalletAPI struct {
```
ElSacramento, line 9: Maybe it's better to rename the file, too.

bigbes: agree

ivanshibitov: yes, done

ledger-core/v2/application/testwalletapi/handlers.go
```diff
@@ -0,0 +1,155 @@
...
+const traceIDField = "traceID"
+
+func create(w http.ResponseWriter, req *http.Request) {
+	traceID := utils.RandTraceID()
+	_, inslog := inslogger.WithTraceField(context.Background(), traceID)
+	inslog.Info("Incoming request: ", req.URL)
+
+	result := map[string]interface{}{
+		"reference":  gen.Reference().String(),
+		traceIDField: traceID,
```
ruz, line 37: please, add here `error: nil` (hope it gets ignored by serialiser), just to indicate how error of the system would be returned in case of any. and in every other map.

ivanshibitov: Done

PR #44, opened 2020-04-13
PLAT-203: functest for wallet/create

ElSacramento:
add test for wallet/create; add api test-utils for tests

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/ledger-core/v2/application/functest/api_test_utils.go
+++ b/ledger-core/v2/application/functest/api_test_utils.go
@@ -33,6 +33,7 @@ const (
 )
 
 func init() {
+	rand.Seed(time.Now().Unix())
 	httpClient = createHTTPClient()
 }
 
@@ -92,7 +93,6 @@ func getURL(path, host, port string) string {
 		host = defaultHost
 	}
 	if port == "" {
-		rand.Seed(time.Now().Unix())
 		port = nodesPorts[rand.Intn(len(nodesPorts))]
 	}
 	res := "http://" + host + ":" + port + path
--- a/ledger-core/v2/application/functest/api_test_utils.go
+++ b/ledger-core/v2/application/functest/api_test_utils.go
@@ -1,0 +1,109 @@
+// Copyright 2020 Insolar Network Ltd.
+// All rights reserved.
+// This material is licensed under the Insolar License version 1.0,
+// available at https://github.com/insolar/assured-ledger/blob/master/LICENSE.md.
+
+// +build functest
+
+package functest
+
+import (
+	"bytes"
+	"encoding/json"
+	"io/ioutil"
+	"math/rand"
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+var (
+	httpClient *http.Client
+	nodesPorts = [5]string{"32301", "32302", "32303", "32304", "32305"}
+)
+
+const (
+	requestTimeout = 30 * time.Second
+	contentType    = "Content-Type"
+
+	defaultHost      = "127.0.0.1"
+	createWalletPath = "/wallet/create"
+)
+
+func init() {
+	httpClient = createHTTPClient()
+}
+
+// createHTTPClient for connection re-use
+func createHTTPClient() *http.Client {
+	client := &http.Client{
+		Transport: &http.Transport{},
+		Timeout:   requestTimeout,
+	}
+
+	return client
+}
+
+// Creates http.Request with all necessary fields.
+func prepareReq(url string, body interface{}) (*http.Request, error) {
+	jsonValue, err := json.Marshal(body)
+	if err != nil {
+		return nil, errors.Wrap(err, "problem with marshaling params")
+	}
+
+	req, err := http.NewRequest(http.MethodPost, url, bytes.NewBuffer(jsonValue))
+	if err != nil {
+		return nil, errors.Wrap(err, "problem with creating request")
+	}
+	req.Header.Set(contentType, "application/json")
+
+	return req, nil
+}
+
+// Executes http.Request and returns response body.
+func doReq(req *http.Request) ([]byte, error) {
+	postResp, err := httpClient.Do(req)
+	if err != nil {
+		return nil, errors.Wrap(err, "problem with sending request")
+	}
+
+	if postResp == nil {
+		return nil, errors.New("response is nil")
+	}
+
+	defer postResp.Body.Close()
+	if http.StatusOK != postResp.StatusCode {
+		return nil, errors.New("bad http response code: " + strconv.Itoa(postResp.StatusCode))
+	}
+
+	body, err := ioutil.ReadAll(postResp.Body)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


ledger-core/v2/application/functest/create_wallet_test.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// Creates wallets and check Reference format in response body.
+func TestCreateWallet(t *testing.T) {
+	for i := 0; i < 100; i++ {
```
ivanshibitov, line 19: Shouldn't  we use for that "count" param in "go test"?

ledger-core/v2/application/functest/create_wallet_test.go
```diff
@@ -0,0 +1,33 @@
...
+			t.Parallel()
+
+			url := getURL(createWalletPath, "", "")
+			rawResp, err := sendAPIRequest(url, nil)
+			require.NoError(t, err, "failed to send request or get response body")
+
+			resp, err := unmarshalCreateWalletResponse(rawResp)
+			require.NoError(t, err, "failed to unmarshal response")
+			require.Nil(t, resp.Err, "problem during execute request")
+			require.Contains(t, resp.Ref, "insolar:", "wrong reference")
```
ivanshibitov, line 30: I would check 'TraceID' also. That it's not empty

ledger-core/v2/application/functest/api_test_utils.go
```diff
@@ -0,0 +1,109 @@
...
+	return body, nil
+}
+
+// Creates full URL for http request.
+func getURL(path, host, port string) string {
+	if host == "" {
+		host = defaultHost
+	}
+	if port == "" {
+		rand.Seed(time.Now().Unix())
```
ivanshibitov, line 95: Might be to do that once? On "test init" procedure

ledger-core/v2/application/functest/create_wallet_test.go
ElSacramento: I think, that `go test -count=100` is better to use for all tests in the package. To run them multiple times, for example in the nightly build.

ruz: I don't think so. However, current implementation is wrong in other way. Loop should be inside of `t.Run`. `t.Parallel()` shouldn't be used in cases when we want to do operations in parallel in a single tests. Purpose of `t.Parallel()` is to tell that some group of tests can be run in parallel and there is no know interference between tests.

ledger-core/v2/application/functest/create_wallet_test.go
ElSacramento: done

ledger-core/v2/application/functest/create_wallet_test.go
ElSacramento: done

ledger-core/v2/application/functest/api_test_utils.go
ElSacramento: done

ledger-core/v2/application/functest/create_wallet_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	t.Run(fmt.Sprintf("count=%d", count), func(t *testing.T) {
+		for i := 0; i < count; i++ {
+			url := getURL(createWalletPath, "", "")
+			rawResp, err := sendAPIRequest(url, nil)
+			require.NoError(t, err, "failed to send request or get response body")
+
+			resp, err := unmarshalCreateWalletResponse(rawResp)
+			require.NoError(t, err, "failed to unmarshal response")
+			require.Nil(t, resp.Err, "problem during execute request")
```
ruz, line 43: I would stick with IsEmpty here as it's not clear what format Err would be in.

ledger-core/v2/application/functest/create_wallet_test.go
```diff
@@ -0,0 +1,48 @@
...
+			countInfo := strings.Split(s, "=")
+			return strconv.Atoi(countInfo[1])
+		}
+	}
+	return 1, nil
+}
+
+// Creates wallets and check Reference format in response body.
+func TestCreateWallet(t *testing.T) {
+	count, err := getCount(os.Args)
```
ruz, line 32: Again, we shouldn't tie logic of tests to arguments of `go test`. Why do we call create API multiple times in a single test? Reasons:

- to repeat a fast operation a few times
- to get more chances to hit pulse change in action
- to hit server side with concurrent requests
- for other reasons you wouldn't get from repeating test using `-count N` argument of `go test`

How do we pick number of repeats? Large enough to increase number of events described above. Small enough to make sure one run doesn't go for ages. Number is test dependent and may change over time.

ElSacramento: I can get a number of working nodes from "node.GetStatus" and multiply it on custom value.
```
status := getStatus(t)
count := 50 * status.WorkingListSize
```
So, the test will depend on network size.

ruz: I'm ok with just constant at this point.

ElSacramento: done

ledger-core/v2/application/functest/create_wallet_test.go
ElSacramento: done

ledger-core/v2/application/functest/create_wallet_test.go
```diff
@@ -0,0 +1,36 @@
...
+		for i := 0; i < count; i++ {
+			url := getURL(createWalletPath, "", "")
+			rawResp, err := sendAPIRequest(url, nil)
+			require.NoError(t, err, "failed to send request or get response body")
+
+			resp, err := unmarshalCreateWalletResponse(rawResp)
+			require.NoError(t, err, "failed to unmarshal response")
+			require.Empty(t, resp.Err, "problem during execute request")
+			require.Contains(t, resp.Ref, "insolar:", "wrong reference")
+			require.NotEmpty(t, resp.TraceID, "traceID mustn't be empty")
```
ita-sammann, line 33: Why there's `require` everywhere? If one fails the whole test will fail? Shouldn't we use `assert`s here? This will allow us to find some flaky errors.

ElSacramento: It is useless to check `resp` if `err` is occurred.  Do you mean to change `require` only for `resp` fields checks?

ita-sammann: Yes, I understand that it's useless. The thing I don't like is that if a single error occurs, the loop is stopped and we won't make all other requests.

ita-sammann: On the second thought, it's not critical for this test, so we can keep it like that.

PR #46, opened 2020-04-13
PLAT-201 new testwallet contract for testing RO

aalekseev-ru:
**- What I did**
add new testwallet contract that will be used for RO functests
**- How I did it**
wrote contract ./v2/application/builtin/contract/testwallet/testwallet.go
ran ./bin/insgocc regen-builtin
checked changed files
**- How to verify it**
wait for new func tests


ledger-core/v2/application/builtin/contract/testwallet/testwallet.go
```diff
@@ -0,0 +1,20 @@
...
+
+package testwallet
+
+import (
+	"github.com/insolar/assured-ledger/ledger-core/v2/logicrunner/builtin/foundation"
+)
+
+// Wallet - basic wallet contract.
+type Wallet struct {
+	foundation.BaseContract
```
ruz, line 14: please add amount/sum as field and add some fixed amount (a constant, for example 1000) to every created wallet.

aalekseev-ru: I planned to add it in https://insolar.atlassian.net/browse/PLAT-205

ita-sammann: If you'll add amount field, pls, use a larger value. Smth like a couple of millions should do. It will allow us to run prolonged tests and not run out of money :)

PR #47, opened 2020-04-13
PLAT-230: golangci linter

sexyhamster:
setup for running golangci linter on all the PRs and new commits to master


sexyhamster:
@ita-sammann I merged two jobs in one, so there won't be double repo fetch. I've discussed this point with @twwlf 
So there will be only one check. lint, units, slow

ledger-core/v2/cmd/heavy-badger/scan.go
```diff
@@ -181,7 +181,7 @@ func (dbs *dbScanner) getAllPulses() (pulses []insolar.Pulse) {
 	for ; err == nil; p, err = pulseStore.Backwards(ctx, p.PulseNumber, 1) {
 		pulses = append(pulses, p)
 	}
-	if err != nil && err != pulsedb.ErrNotFound {
+	if err != pulsedb.ErrNotFound {
```
ElSacramento, line 184: If err will be != ErrNotFound , err = SomeError for example, fatalf will not happen, is it ok?

PR #56, opened 2020-04-14
PLAT-240: tracked files checker

ita-sammann:
This step checks that no tracked files have changed in the repository after `make vendor pre-build build`

 * Added script that checks if the working directory is clean
 * Added this checker to GH Actions after the build step

### DIFF
--- a/.github/workflows/common.yaml
+++ b/.github/workflows/common.yaml
@@ -33,6 +33,9 @@ jobs:
           go mod download
           make vendor pre-build build
         working-directory: ${{env.GOPATH}}/src/github.com/insolar/assured-ledger/ledger-core/v2
+      - name: Check if working directory is clean after build
+        run: scripts/gitstatus.sh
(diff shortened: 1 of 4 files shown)


.github/workflows/common.yaml
```diff
@@ -33,6 +33,9 @@ jobs:
           go mod download
           make vendor pre-build build
         working-directory: ${{env.GOPATH}}/src/github.com/insolar/assured-ledger/ledger-core/v2
+      - name: Check if working directory is clean after build
+        run: scripts/gitstatus.sh
+        working-directory: ${{env.GOPATH}}/src/github.com/insolar/assured-ledger/ledger-core/v2
```
sexyhamster, line 38: I wonder  could we set 
```if: always()```
for this and lint rules, so we run tests even if linting/generate fails. To be able to faster see all the problems

ita-sammann: Let's wait for the first situation where we really need it

PR #60, opened 2020-04-15
PLAT-134: headless_network binary removed, insolard CLI for headless network

AndreyBronin:
**- What I did**

- remove insolard_headless binary
- begin implementing correct CLI commands and subcommands (for task PLAT-182)
- using correct CLI commands in launchnet script


ledger-core/v2/scripts/insolard/launchnet.sh
```diff
@@ -491,8 +492,18 @@ echo "log: ${DISCOVERY_NODE_LOGS}1/output.log"
 echo "start discovery nodes ..."
 for i in `seq 2 $NUM_DISCOVERY_NODES`
 do
+    ROLE=""
```
AndreyBronin, line 495: the workaround here is copied from https://github.com/insolar/insolar-scripts

AndreyBronin:
> lgfm, but can you actualize README at https://github.com/insolar/assured-ledger/tree/master/ledger-core/v2 and add there examples of running network in different configurations?

this PR is only for headless network mode. Other modes still not ready to use

PR #61, opened 2020-04-15
PLAT-238: Don't change pulse in insolar net

ivanshibitov:
…luanch pulsar with option --one-shot. 3. Increase PulseWatchdogTimeout up to 1 month

**- What I did**
1. Func tests run luanchnet.sh without pulsar. 
2. Func tests run pulsar with option --one-shot. 
3. Increase PulseWatchdogTimeout up to 1 month to prevent crash of nodes

**- How I did it**

**- How to verify it**

### DIFF
--- a/ledger-core/v2/application/testutils/launchnet/launchnet.go
+++ b/ledger-core/v2/application/testutils/launchnet/launchnet.go
@@ -24,7 +24,7 @@ import (
 	"testing"
 	"time"
 
-	yaml "gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v2"
 
 	"github.com/pkg/errors"
 
@@ -359,7 +359,7 @@ func stopInsolard() error {
 	return nil
 }
 
-func waitForNet() error {
+func waitForNetworkState(state insolar.NetworkState) error {
 	numAttempts := 270
 	// TODO: read ports from bootstrap config
 	ports := []string{
@@ -377,6 +377,7 @@ func waitForNet() error {
 	}
 	numNodes := len(ports)
 	currentOk := 0
+	fmt.Println("Waiting for Network state: ", state.String())
 	for i := 0; i < numAttempts; i++ {
 		currentOk = 0
 		for _, port := range ports {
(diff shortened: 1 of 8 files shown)


ledger-core/v2/application/testutils/launchnet/launchnet.go
```diff
@@ -404,6 +404,34 @@ func waitForNet() error {
...
+	output, err := pulsarCmd.CombinedOutput()
+	fmt.Println("Pulsar launch output: ", string(output))
+
+	return errors.Wrap(err, "failed to launch pulsar")
+}
+
+func waitForNet() error {
+	err := waitForNetworkState(insolar.WaitPulsar)
+	if err != nil {
+		return errors.Wrap(err, "Can't for NetworkState "+insolar.WaitPulsar.String())
```
ElSacramento, line 423: Can't get NetworkState? Can't wait for state? what can't do?

ledger-core/v2/application/testutils/launchnet/launchnet.go
```diff
@@ -404,6 +404,34 @@ func waitForNet() error {
...
+	}
+
+	err = runPulsar()
+	if err != nil {
+		return errors.Wrap(err, "Can't run pulsar")
+	}
+
+	err = waitForNetworkState(insolar.CompleteNetworkState)
+	if err != nil {
+		return errors.Wrap(err, "Can't for NetworkState "+insolar.WaitPulsar.String())
```
ElSacramento, line 433: `insolar.CompleteNetworkState.String()`

ledger-core/v2/cmd/pulsard/main.go
```diff
@@ -104,6 +106,18 @@ func main() {
...
+
+	if params.oneShot {
+		nextPulseNumber := pulse.OfNow()
+		err := server.Send(ctx, nextPulseNumber)
+		if err != nil {
+			panic(err)
+		}
+		// it's required since pulse is sent in goroutine
+		time.Sleep(time.Second * 10)
+		return
```
ElSacramento, line 118: I see `err = cm.Stop(ctx)` in defer. Do we need to stop component manager in one-shot run?

ivanshibitov: There is no any action we can do if it's not stopped.
And it's the end of the program.
And code is shorter.
That is why I didn't call Stop().
But, for perfectness - yes.

PR #63, opened 2020-04-15
PLAT-229: check code coverage using GitHub Actions

afiskon:
Check code coverage using GitHub Actions.

### DIFF
--- a/.github/workflows/common.yaml
+++ b/.github/workflows/common.yaml
@@ -41,3 +41,11 @@ jobs:
       - name: Run unit and slow tests
         run: make test_slow
         working-directory: ${{env.GOPATH}}/src/github.com/insolar/assured-ledger/ledger-core/v2
+      - name: Check code coverage
+        run: make test_with_coverage
+        working-directory: ${{env.GOPATH}}/src/github.com/insolar/assured-ledger/ledger-core/v2
+      - name: Upload coverage report
+        uses: actions/upload-artifact@v1
+        with:
+          name: coverage-report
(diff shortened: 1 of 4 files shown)


afiskon:
Seems to work. All checks were passed, coverage.html was attached in Artifacts.

sexyhamster:
hmmm how to see the report? :octocat:

afiskon:
@sexyhamster  Open Checks tab, find Artifacts on the left, download the archive.

sexyhamster:
that is really hard to control on review >_<

afiskon:
@sexyhamster there is no need to control it during the code review. Total coverage is checked automatically. Currently it's 39.1%, if it drops below 39% the check will not pass.

sexyhamster:
maybe we can do something like https://github.com/jandelgado/golang-ci-template-github-actions ? so it not only check but actually comment

afiskon:
@sexyhamster Coversall looks nice, however I thought the whole idea was not to depend on extra 3rd party service since we had a problem with such services before.

sexyhamster:
hmm, I had no problem with service itself, I was concerned, that it showed random numbers on PR :( Maybe we were doing something wrong. But I really like visual comment, that can control not only threshold for coverage, but coverage changes with each PR. 

Coverage will be run manually in check, but service can actually create comment with diff from master. We could do it without service, I think.

What i want - is to have comment - merging this PR will increase/decrease coverage by N%.

afiskon:
Ok after discussion with @ita-sammann turned out this is not exactly what we wanted. I will consider options with https://coveralls.io/ and https://codecov.io/

PR #67, opened 2020-04-16
PLAT-245: remove old func tests

sexyhamster:
Removed all the redundant func tests that were testing Pl…atform 1.0 behaviour


ledger-core/v2/application/functest/get_info_test.go
```diff
@@ -1,22 +0,0 @@
...
-
-package functest
-
-import (
-	"testing"
-
-	"github.com/stretchr/testify/require"
-)
-
-func TestGetInfo(t *testing.T) {
```
ElSacramento, line 16: It works now (jfyi).

ruz: @ElSacramento, do we need this test? It's not clear from your comment.

ElSacramento: We have discussed on one of the team meetings, that it's ok to delete not working tests, but this test works. I'm not sure that it's useful now for platform 2.0, but while it works maybe we can leave it? Just a thought.

sexyhamster: We discussed in Plat 246, that we don't need this contract, so it will be removed in #69

ruz: took closer look. agree.

PR #69, opened 2020-04-16
PLAT-246: remove mainnet contracts

ivanshibitov:
**- What I did**
1. Remove all contracts, their wrappers and proxies excluding testWallet( it's used in integration tests of virtual )
2. Remove all contracts from genesis
3. Remove part of API which works with contracts

**- How I did it**

**- How to verify it**


ledger-core/v2/application/testutils/launchnet/launchnet.go
```diff
@@ -3,6 +3,8 @@
 // This material is licensed under the Insolar License version 1.0,
 // available at https://github.com/insolar/assured-ledger/blob/master/LICENSE.md.
 
+// +build ZHEP_EBRILO
```
bigbes, line 6: remove it pls

ledger-core/v2/application/builtin/contract/testwallet/testwallet.go
```diff
@@ -6,53 +6,54 @@
 package testwallet
 
 import (
-	"errors"
-	"fmt"
+       "errors"
```
bigbes, line 9: It's strange, but in that file all tabs were replaced with spaces. Why?

ivanshibitov: hmmm...

PR #74, opened 2020-04-17
PLAT-260: functests on CI

ita-sammann:
Added functests step to `common.yaml` workflow

### DIFF
--- a/ledger-core/v2/Makefile
+++ b/ledger-core/v2/Makefile
@@ -51,6 +51,7 @@ clean: ## run all cleanup tasks
 	go clean $(ALL_PACKAGES)
 	rm -f $(COVERPROFILE)
 	rm -rf $(BIN_DIR)
+	rm -rf $(ARTIFACTS_DIR)
 	./scripts/insolard/launchnet.sh -l
 
 .PHONY: install-build-tools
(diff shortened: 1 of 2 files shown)


ledger-core/v2/Makefile
```diff
@@ -51,6 +51,7 @@ clean: ## run all cleanup tasks
 	go clean $(ALL_PACKAGES)
 	rm -f $(COVERPROFILE)
 	rm -rf $(BIN_DIR)
+	rm -rf $(ARTIFACTS_DIR)
```
ElSacramento, line 54: I can't find using `make clean` in github actions.

ita-sammann: It's used in `launchnet.sh`

ElSacramento: In `rebuild_binaries`? but func is unused :(

ita-sammann: Hmm, yes, really. But nevertheless, I use it all the time manually :)

ElSacramento: Maybe it's time to add it in `test_func` step or in `launchnet` :)

ita-sammann: To run tests on a clean environment I run `make` (there is `clean` target in it) and then `make test_func`

PR #75, opened 2020-04-20
PLAT-215: Implemented transfer API call to testwallet contract

dgovorukhin:


ledger-core/v2/application/testwalletapi/handlers.go
```diff
@@ -140,6 +140,53 @@ func (s *TestWalletServer) Transfer(w http.ResponseWriter, req *http.Request) {
 		Error:   "",
 	}
 	defer func() { s.mustWriteResult(w, result) }()
+
+	fromRef, err := insolar.NewReferenceFromString(params.From)
+	if err != nil {
+		result.Error = fmt.Sprintf("Failed to create reference from string (%s), sender reference", params.From)
```
sexyhamster, line 146: probably we should log original error

ledger-core/v2/application/testwalletapi/handlers.go
```diff
@@ -140,6 +140,53 @@ func (s *TestWalletServer) Transfer(w http.ResponseWriter, req *http.Request) {
...
+	}
+
+	toRef, err := insolar.NewReferenceFromString(params.To)
+	if err != nil {
+		result.Error = fmt.Sprintf("Failed to create reference from string (%s), receiver reference", params.To)
+		return
+	}
+
+	//TODO how to serialize multiple method parameters?
+	transferParams := make([]interface{}, 2)
```
sexyhamster, line 157: make([]interface{}, 0, 2)

sexyhamster: or even 
```insolar.Serialize([]interface{}{toRef, params.Amount})```

File: ledger-core/v2/application/testwalletapi/handlers.go [hunk 3]
```diff
@@ -140,6 +140,53 @@ func (s *TestWalletServer) Transfer(w http.ResponseWriter, req *http.Request) {
...
+		return
+	}
+
+	var contractCallErr *foundation.Error
+	err = foundation.UnmarshalMethodResultSimplified(walletRes.ReturnArguments, &contractCallErr)
+	switch {
+	case err != nil:
+		result.Error = errors.Wrap(err, "Failed to unmarshal response").Error()
+	case contractCallErr != nil:
+		result.Error = contractCallErr.S
```
sexyhamster, line 186: I'd prefer to use method contractCallErr.Error()

ledger-core/v2/application/testwalletapi/handlers.go
```diff
@@ -140,6 +140,53 @@ func (s *TestWalletServer) Transfer(w http.ResponseWriter, req *http.Request) {
...
+	}
+
+	walletRes, err := s.runWalletRequest(ctx, walletReq)
+	if err != nil {
+		result.Error = err.Error()
+		return
+	}
+
+	var contractCallErr *foundation.Error
+	err = foundation.UnmarshalMethodResultSimplified(walletRes.ReturnArguments, &contractCallErr)
```
sexyhamster, line 181: it is already *foundation.Error, so you are passing **foundation.Error, instead of *foundation.Error

ledger-core/v2/application/testwalletapi/handlers.go
sexyhamster: I think it would be useful in error cases

dgovorukhin: Fixed

ledger-core/v2/application/testwalletapi/handlers.go
dgovorukhin: Fixed

ledger-core/v2/application/testwalletapi/handlers.go
dgovorukhin: Fixed

ledger-core/v2/application/testwalletapi/handlers.go
dgovorukhin: Fixed

dgovorukhin: As I can see in details an implementation UnmarshalMethodResultSimplified, method expected returns[i].(**Error) to find the needed argument

sexyhamster: my bad

ledger-core/v2/application/testwalletapi/handlers.go
```diff
@@ -139,23 +139,32 @@ func (s *TestWalletServer) Transfer(w http.ResponseWriter, req *http.Request) {
...
+			logger.Error(result.Error)
+		}
+		s.mustWriteResult(w, result)
+	}()
 
 	fromRef, err := insolar.NewReferenceFromString(params.From)
 	if err != nil {
-		result.Error = err.Error()
+		result.Error = throw.W(err,
+			fmt.Sprintf("Failed to create reference from string (%s), sender reference", params.From), nil,
```
sexyhamster, line 152: better logging usage: it's not very clear what usage is for, but as far as I see it should be like that
```
 		result.Error = throw.W(err,  "failted to create reference from string", struct{ From string }{ From: params.From }
```

dgovorukhin: Fixed)

PR #76, opened 2020-04-20
PLAT-206: Implemented getBalance and addAmount API call to testwallet

dgovorukhin:


ledger-core/v2/application/testwalletapi/handlers.go
```diff
@@ -175,7 +175,43 @@ func (s *TestWalletServer) GetBalance(w http.ResponseWriter, req *http.Request)
 		TraceID: traceID,
 		Error:   "",
 	}
+
 	defer func() { s.mustWriteResult(w, result) }()
```
ivanshibitov, line 179: Don't we want to log error if it occurs?
Something like:
`
defer func() { 
if len(  result.Error) != 0{
log( result.Error )
}
s.mustWriteResult(w, result) 
}()
`

ledger-core/v2/application/testwalletapi/handlers.go
```diff
@@ -175,7 +175,43 @@ func (s *TestWalletServer) GetBalance(w http.ResponseWriter, req *http.Request)
...
+
+	var (
+		amount          uint32
+		contractCallErr *foundation.Error
+	)
+
+	err = foundation.UnmarshalMethodResultSimplified(walletRes.ReturnArguments, &amount, &contractCallErr)
+	switch {
+	case err != nil:
+		result.Error = errors.Wrap(err, "Failed to unmarshal response").Error()
```
ivanshibitov, line 209: You wrap error here and didn't do that on line 197. Why?

ledger-core/v2/application/testwalletapi/testwalletserver.go
```diff
@@ -29,7 +29,9 @@ type TestWalletServer struct {
 }
 
 const (
-	create = "New"
+	create     = "New"
```
ivanshibitov, line 32: Give comment and rename this constant. It's not clear, what there are about

dgovorukhin: Is it more clear for you?
// Aliases VCallRequest.CallSiteMethod for call testwallet.Wallet contract methods
const (
	create     = "New"     // Create wallet
	getBalance = "Balance" // Get wallet balance
	addAmount  = "Accept"  // Add money to wallet
)

ledger-core/v2/application/testwalletapi/handlers.go
sexyhamster: the error will go to functest, if it is not expected, it will be logged/checked there. In Border contracts we will log contract output for sure

ledger-core/v2/application/testwalletapi/handlers.go
dgovorukhin: fixed

ledger-core/v2/application/testwalletapi/handlers.go
ivanshibitov: Why it's not expected to be logged here?
If there is any problem, I just grep traceID in log and see all path of request

ledger-core/v2/application/testwalletapi/handlers.go
```diff
@@ -175,7 +175,50 @@ func (s *TestWalletServer) GetBalance(w http.ResponseWriter, req *http.Request)
 		TraceID: traceID,
 		Error:   "",
 	}
-	defer func() { s.mustWriteResult(w, result) }()
+
+	defer func() {
+		if len(result.Error) != 0 {
+			logger.Error(result.Error)
```
ElSacramento, line 181: I think we must return her and don't write `result`.

dgovorukhin: why? Here we try to log the error and send a response with an error. What's the problem?

ElSacramento: I think that in result could be wrong data, that is not necessary to return, but this error is a part of result, so that's ok.

PR #84, opened 2020-04-22
PLAT-208 sending VStateRequest on Init object

aalekseev-ru:
**- What I did**
add wait for object state
**- How I did it**
add new step
**- How to verify it**
wait for tests


ledger-core/v2/virtual/object/object.go
```diff
@@ -91,16 +112,57 @@ func (sm *SMObject) Init(ctx smachine.InitializationContext) smachine.StateUpdat
 	sm.ImmutableExecute = smsync.NewSemaphore(30, "immutable calls").SyncLink()
 	sm.MutableExecute = smsync.NewSemaphore(1, "mutable calls").SyncLink() // TODO here we need an ORDERED queue
 
+	sm.alreadySendVStateRequest = false
```
sexyhamster, line 115: default bool value is false already

sexyhamster: maybe stateRequestSent would be better name

aalekseev-ru: fixed

sexyhamster:
Looks OK to me, but tests are required

ledger-core/v2/virtual/object/object.go
```diff
@@ -95,12 +111,45 @@ func (sm *SMObject) Init(ctx smachine.InitializationContext) smachine.StateUpdat
...
+
+	switch sm.initReason {
+	case InitReasonCTConstructor:
+		return ctx.Jump(sm.stepReadyToWork)
+	case InitReasonCTMethod:
+		return ctx.Jump(sm.stepGetObjectState) // TODO PLAT-294 - we will jump to stepWaitState
+	case InitReasonVStateReport:
+		return ctx.Jump(sm.stepWaitState)
+	default:
+		panic("Not implemented")
```
bigbes, line 123: panic(throw.IllegalValue())

ledger-core/v2/virtual/object/object.go
```diff
@@ -95,12 +111,45 @@ func (sm *SMObject) Init(ctx smachine.InitializationContext) smachine.StateUpdat
...
+// we get CallMethod but we have no object data
+// we need to ask previous executor
+func (sm *SMObject) stepGetObjectState(ctx smachine.ExecutionContext) smachine.StateUpdate {
+	msg := payload.VStateRequest{
+		Callee:           sm.Reference,
+		RequestedContent: payload.RequestLatestDirtyState,
+	}
+
+	goCtx := ctx.GetContext()
+	sm.messageSender.PrepareNotify(ctx, func(svc messagesender.Service) {
```
bigbes, line 136: Sure, `PrepareNotify().Send()`is called in context of current SM, but try to avoid direct access to sm memory in all adapter callbacks.
For readability, please, extract `sm.pulseSlot.PulseData().PrevPulseNumber()` to separate variable.
And also i'm not sure that we should use `.PrevPulseNumber()` there, since pulse can be skipped. We need somehow to get previous pulse from storage.

ledger-core/v2/virtual/integration/method_api_test.go
```diff
@@ -84,7 +84,7 @@ func TestVirtual_Method_API(t *testing.T) {
 	})
 
 	t.Run("Transfer", func(t *testing.T) {
-		t.Skip("NYI")
+		t.Skip("Not implemented yet")
```
bigbes, line 87: whats for?

ledger-core/v2/virtual/object/object.go
ruz, line 123: this is wrong. you should use method from package `throw.NotImplemented`. However, in this particular case error should be `IlleagalValue`.

ledger-core/v2/virtual/integration/method_api_test.go
aalekseev-ru: what means NYI?

ledger-core/v2/virtual/object/object.go
aalekseev-ru: fixed

ledger-core/v2/virtual/integration/method_api_test.go
bigbes: Not Yet Implemented, commonly used

PR #87, opened 2020-04-23
PLAT-276: counters of pending execution

MikhailZhukov:
**- What I did**
Add counters of pending execution to object state

**- How I did it**
Add fields to struct `Info`

**- How to verify it**

This counters must be filled when processing VStateReport (PLAT-210)

### DIFF
--- a/ledger-core/v2/virtual/object/object.go
+++ b/ledger-core/v2/virtual/object/object.go
@@ -14,9 +14,6 @@ import (
 	"github.com/insolar/assured-ledger/ledger-core/v2/virtual/descriptor"
 )
 
-// Stores information about Executor (key) and its pending execution requests (value)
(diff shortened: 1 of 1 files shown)


ledger-core/v2/virtual/object/object.go
```diff
@@ -14,6 +14,9 @@ import (
 	"github.com/insolar/assured-ledger/ledger-core/v2/virtual/descriptor"
 )
 
+// Stores information about Executor (key) and its pending execution requests (value)
```
ruz, line 17: I don't think it's proper way to organise this table. It would be hard to do de-duplication checks. I, also, think that this is out of scope for this task.

MikhailZhukov: Remove this queue from this PR

PR #89, opened 2020-04-23
PLAT-204: implement unordered terminal method calls

bigbes:


ledger-core/v2/logicrunner/preprocessor/templates/wrapper.go.tpl
```diff
@@ -263,7 +263,10 @@ func Initialize() insolar.ContractWrapper {
 		GetPrototype: INSMETHOD_GetPrototype,
 		Methods: insolar.ContractMethods{
 			{{ range $method := .Methods -}}
-					"{{ $method.Name }}": INSMETHOD_{{ $method.Name }},
+					"{{ $method.Name }}": insolar.ContractMethod{
+					    Func: INSMETHOD_{{ $method.Name }},
+					    Unordered: {{ $method.Immutable }},
```
aalekseev-ru, line 268: It seems kinda strange. Do we now have 2 different types of call - unordered AND immutable? As I know, now we replace what we used to call 'immutable' with 'Unordered'. So I think we need to remove 'Immutable' from code base. And it is strange that we change the way wrapper is generated.

ledger-core/v2/runner/executor/builtin/builtin.go
```diff
@@ -86,12 +87,26 @@ func (b *Runner) CallMethod(
 	}
 	contract := b.CodeRegistry[contractName]
 
-	methodFunc, ok := contract.Methods[method]
+	methodFuncObject, ok := contract.Methods[method]
```
aalekseev-ru, line 90: If we rename it to methodObject we won't use something like return methodFunc.Func(data, args) (see above)

ledger-core/v2/logicrunner/builtin/builtin.go
```diff
@@ -121,5 +121,5 @@ func (b *BuiltIn) CallMethod(ctx context.Context, callCtx *insolar.LogicCallCont
 		return nil, nil, errors.New("failed to find contracts method")
 	}
 
-	return methodFunc(data, args)
+	return methodFunc.Func(data, args)
```
aalekseev-ru, line 124: maybe rename to methodObject?

ledger-core/v2/virtual/integration/method_test.go
```diff
@@ -159,6 +162,104 @@ func TestVirtual_Method_WithoutExecutor(t *testing.T) {
 	}
 }
 
+func TestVirtual_Method_WithoutExecutor_Immutable(t *testing.T) {
```
aalekseev-ru, line 165: why it is still 'immutable'?

ledger-core/v2/logicrunner/preprocessor/templates/wrapper.go.tpl
bigbes: If it's need to be done - it should be different PR

ledger-core/v2/runner/executor/builtin/builtin.go
bigbes: sure

ledger-core/v2/logicrunner/builtin/builtin.go
bigbes: Old codebase, just supported compilation

ledger-core/v2/virtual/integration/method_test.go
bigbes: sure

ledger-core/v2/logicrunner/builtin/builtin.go
ruz: https://insolar.atlassian.net/browse/PLAT-298

ledger-core/v2/logicrunner/preprocessor/templates/wrapper.go.tpl
ruz: https://insolar.atlassian.net/browse/PLAT-296

PR #92, opened 2020-04-24
Plat 278 update pending counters in incoming requests

MikhailZhukov:
**- What I did**
Increment pending counters when processing new message

**- How I did it**
Add increment code to SMExecute.stepWaitObjectReady

**- How to verify it**

### DIFF
--- a/ledger-core/v2/conveyor/smachine/api_shared_link.go
+++ b/ledger-core/v2/conveyor/smachine/api_shared_link.go
@@ -8,22 +8,9 @@ package smachine
 import (
 	"fmt"
 	"reflect"
-
-	"github.com/insolar/assured-ledger/ledger-core/v2/vanilla/throw"
 )
 
 type SharedDataFunc func(interface{}) (wakeup bool)
-
-func NewUnboundSharedData(data interface{}) SharedDataLink {
-	if data == nil {
-		panic(throw.IllegalValue())
-	}
-
-	return SharedDataLink{
-		data:  data,
-		flags: ShareDataUnbound,
-	}
-}
 
 // Link to a data shared by a slot.
 // This link can live longer than its origin.
--- a/ledger-core/v2/conveyor/smachine/api_shared_link.go
+++ b/ledger-core/v2/conveyor/smachine/api_shared_link.go
@@ -8,9 +8,22 @@ package smachine
 import (
 	"fmt"
 	"reflect"
+
+	"github.com/insolar/assured-ledger/ledger-core/v2/vanilla/throw"
 )
 
 type SharedDataFunc func(interface{}) (wakeup bool)
+
+func NewUnboundSharedData(data interface{}) SharedDataLink {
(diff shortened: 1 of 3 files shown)


ledger-core/v2/conveyor/smachine/api_shared_link.go
```diff
@@ -8,10 +8,23 @@ package smachine
 import (
 	"fmt"
 	"reflect"
+
+	"github.com/insolar/assured-ledger/ledger-core/v2/vanilla/throw"
 )
 
 type SharedDataFunc func(interface{}) (wakeup bool)
 
+func NewUnboundSharedData(data interface{}) SharedDataLink {
```
ivanshibitov, line 17: What for?

ivanshibitov: I mean this function is not used anywhere

MikhailZhukov: This function create SharedDataLink which sometimes need for tests. But unit test was deleted (it does not work). Do i must remove this function too?

MikhailZhukov: Yes, it does not used right now. I can remove it. But maybe it useful for future tests?

MikhailZhukov: remove this code

ledger-core/v2/virtual/execute/execute.go
```diff
@@ -145,6 +145,12 @@ func (s *SMExecute) stepWaitObjectReady(ctx smachine.ExecutionContext) smachine.
 		semaphoreOrdered = state.MutableExecute
 		semaphoreUnordered = state.ImmutableExecute
 
+		if isOrdered {
```
ivanshibitov, line 148: Lower, There is following code
if ctx.AcquireForThisStep(semaphoreReadyToWork).IsNotPassed() {
		return ctx.Sleep().ThenRepeat()
}
it means that if we can't acquire the lock we will repeat this step and again increment this numbers

MikhailZhukov: Fix multiple increment on retries

ledger-core/v2/virtual/object/object.go
```diff
@@ -30,10 +30,23 @@ type Info struct {
...
-	PotentialMutablePendingCount   uint8
+	ActiveImmutablePendingCount           uint8
+	ActiveMutablePendingCount             uint8
+	PotentialImmutablePendingCount        uint8
+	PotentialMutablePendingCount          uint8
+	potentialPendingCounterWasIncremented bool
+}
+
+func (i *Info) IncrementPotentialPendingCounterOnce(isOrdered bool) {
+	if i.potentialPendingCounterWasIncremented {
```
bigbes, line 41: I don't get that logic, so we can increment only once?

PR #96, opened 2020-04-25
NOISSUE: cleanup old messages and code

cyraxred:
**- What I did**
- started from old messages then removed everything that depends on them


ledger-core/v2/insolar/pulse/memory_test.go
```diff
@@ -24,11 +25,11 @@ func TestNodeStorage_ForPulseNumber(t *testing.T) {
...
-	t.Run("returns error when no Pulse", func(t *testing.T) {
-		res, err := storage.ForPulseNumber(ctx, gen.PulseNumber())
-		assert.Equal(t, ErrNotFound, err)
-		assert.Equal(t, insolar.Pulse{}, res)
-	})
+	// t.Run("returns error when no Pulse", func(t *testing.T) {
+	// 	res, err := storage.ForPulseNumber(ctx, gen.PulseNumber())
+	// 	assert.Equal(t, ErrNotFound, err)
+	// 	assert.Equal(t, insolar.Pulse{}, res)
+	// })
```
afiskon, line 32: FYI generally we don't keep commented code. It can always be recovered using VCS if necessary. I believe it somewhere in our code style.

ledger-core/v2/ledger/object/record_db.go
```diff
@@ -10,20 +10,18 @@ import (
...
 	"github.com/insolar/assured-ledger/ledger-core/v2/insolar"
-	"github.com/insolar/assured-ledger/ledger-core/v2/insolar/record"
 	"github.com/insolar/assured-ledger/ledger-core/v2/insolar/store"
 	"github.com/insolar/assured-ledger/ledger-core/v2/instrumentation/inslogger"
 )
 
 // RecordDB is a DB storage implementation. It saves records to disk and does not allow removal.
 type RecordDB struct {
-	batchLock sync.Mutex
+	// batchLock sync.Mutex
```
afiskon, line 24: ditto - there is no need to keep this line here.

afiskon:
On second thought please fix conflicts first.

ledger-core/v2/insolar/pulse/memory_test.go
cyraxred: Well, I like to revisit code multiple times, and comments are more efficient. VCS is only good when you know where and which version to look at. It is very inefficient otherwise.

I've restored this test

ledger-core/v2/ledger/object/record_db.go
cyraxred: this gonna be removed soon

PR #97, opened 2020-04-25
PLAT-302: without heavy and light

ruz:


ledger-core/v2/network/gateway/base.go
```diff
@@ -273,32 +274,21 @@ func (g *Base) checkCanAnnounceCandidate(ctx context.Context) error {
 	// 		NB: announcing in WaitConsensus state is *NOT* allowed
```
AndreyBronin, line 274: rewrite steps of the algorithm in comments, replace

-  "node is heavy" with "JoinAssistant"
- "when heavy node found" with "when JoinAssistant node found"

PR #103, opened 2020-04-27
PLAT-211: VStateUnavailable handler

ivanshibitov:
**- What I did**
1. Add message VStateUnavailable
2. Add handler for 1.
3. Simple test that just send that message

**- How I did it**

**- How to verify it**


ledger-core/v2/virtual/request/vstateunavailable.go
```diff
@@ -0,0 +1,85 @@
...
+	objectRef := s.Payload.Reference
+	sharedObjectState, ok := catalog.TryGet(ctx, objectRef)
+	if !ok {
+		ctx.Log().Error(noObjectErrorMsg{Reference: objectRef.String()}, nil)
+		return ctx.Stop()
+	}
+
+	setStateFunc := func(state *object.SharedState) {
+		if state.IsReady() {
+			panic(throw.IllegalState())
```
bigbes, line 63: Are you sure it's `IllegalState`, and not "strange things happened, just ignore it"?

ledger-core/v2/virtual/request/vstatereport.go
```diff
@@ -51,19 +51,18 @@ func (s *SMVStateReport) stepProcess(ctx smachine.ExecutionContext) smachine.Sta
 	objectRef := incomingObjectState.Reference
 	sharedObjectState := catalog.GetOrCreate(ctx, objectRef, object.InitReasonVStateReport)
 
-	setStateFunc := func(data interface{}) (wakeup bool) {
-		state := data.(*object.SharedState)
-		if state.Descriptor() == nil {
+	setStateFunc := func(state *object.SharedState) {
+		if !state.IsReady() {
```
bigbes, line 55: Not sure that'll work correctly with "Inactive" object, if we'll receive "Missing" or other state.

ledger-core/v2/virtual/request/vstateunavailable.go
ivanshibitov: I think it's more about what we do with such situations right now.
Possibly yes, it's better just to log it.
I'm not sure

ledger-core/v2/virtual/request/vstatereport.go
bigbes: just sort state:
```
type State int32

const (
	NotReady State = iota
	Missing = 1
	HasState  = 2
	Inactive = 3
)
```

And use ">" for comparing applicable jumps, I guess

ledger-core/v2/virtual/object/object.go
```diff
@@ -21,6 +21,15 @@ import (
 	"github.com/insolar/assured-ledger/ledger-core/v2/virtual/descriptor"
 )
 
+type State int32
+
+const (
```
bigbes, line 26: use iota, pls

ledger-core/v2/virtual/integration/stateunavailable_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+func makeVStateUnavailableEvent(t *testing.T, ref insolar.Reference) *statemachine.DispatcherMessage {
+	payLoadMeta := &payload.VStateUnavailable{
+		Reference: ref,
+		Reason:    payload.Inactive,
+	}
+	return makeDispatcherMessage(t, payLoadMeta)
+}
+
+func TestVirtual_VStateUnavailable_NoSuchObject(t *testing.T) {
```
bigbes, line 30: I'd suggest adding two more tests for all possible "Reasons", or is it out of scope of that task?

ledger-core/v2/virtual/object/object.go
```diff
@@ -147,7 +166,7 @@ func (sm *SMObject) stepGetObjectState(ctx smachine.ExecutionContext) smachine.S
 }
 
 func (sm *SMObject) stepWaitState(ctx smachine.ExecutionContext) smachine.StateUpdate {
-	if sm.SharedState.descriptor != nil {
+	if sm.IsReady() {
```
bigbes, line 169: Different logic for Inactive, probably? Or is it out of scope?

ledger-core/v2/virtual/integration/stateunavailable_test.go
ivanshibitov: I will add

ledger-core/v2/virtual/request/vstateunavailable.go
bigbes: Or applying suggested logic from the `setStateFunc` from VStateReport?

ledger-core/v2/virtual/object/object.go
ivanshibitov: I think it's not in this task

ledger-core/v2/virtual/execute/execute.go
```diff
@@ -277,13 +277,15 @@ func (s *SMExecute) stepSaveNewObject(ctx smachine.ExecutionContext) smachine.St
...
 			state.Info.SetDescriptor(&prototype, memory)
+			state.SetState(object.HasState)
 		}
 
 	case insolar.RequestSideEffectAmend:
 		_, prototype, memory = executionNewState.Amend()
 		action = func(state *object.SharedState) {
 			ctx.Log().Trace("applying new state")
 			state.Info.SetDescriptor(&prototype, memory)
+			state.SetState(object.HasState)
```
afiskon, line 288: Looks a little error-prone. Maybe this change should be made automatically in Set* methods?

ledger-core/v2/virtual/object/object.go
```diff
@@ -34,6 +43,16 @@ type Info struct {
 	ActiveMutablePendingCount      uint8
 	PotentialImmutablePendingCount uint8
 	PotentialMutablePendingCount   uint8
+
+	objectState State
+}
+
+func (i *Info) IsReady() bool {
+	return i.objectState != NotReady
```
afiskon, line 51: Maybe "Unknown" would be a better name that "NotReady" in this context?

ledger-core/v2/virtual/execute/execute.go
ivanshibitov: Yes. But
For example in VStateReport we will copy a lot of fields from incoming message to 'state'.
And object is ready only all of them are copied. In this case it's not clear hot to that automatically

ledger-core/v2/virtual/request/vstateunavailable.go
afiskon: I suggest to keep panic here if the author believes that it doesn't suppose to happen. Maybe `throw.Impossible()` would be even better if this is a case.

ledger-core/v2/virtual/request/vstatereport.go
ivanshibitov: O_O

ledger-core/v2/virtual/object/object.go
ivanshibitov: + I think logic in SMObject will be the same.
It will be changed in places, where SMObject is used

ledger-core/v2/virtual/integration/stateunavailable_test.go
```diff
@@ -0,0 +1,48 @@
+// Copyright 2020 Insolar Network Ltd.
+// All rights reserved.
+// This material is licensed under the Insolar License version 1.0,
+// available at https://github.com/insolar/assured-ledger/blob/master/LICENSE.md.
+
+package small
```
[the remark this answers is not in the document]
ivanshibitov: All integration tests are in this packet.
I can move all of them to another.
Or we can do it in separate task.
@bigbes What do you think?

ledger-core/v2/virtual/request/vstatereport.go
ivanshibitov: I think its not good idea to rely on order of these constants. It might be changed accidentally. 
But yes,  question about different states in VStateUnavailable and in SMObject is opened.
Right now it's not clear what to do

ledger-core/v2/virtual/integration/stateunavailable_test.go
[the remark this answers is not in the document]
bigbes: It was in folder "small" before, i suggest we should do that in separate PR

ledger-core/v2/virtual/request/vstatereport.go
bigbes: Relying on order of constants is a common practice, that's why one shouldn't change order of constants without giving it a second thought.

ledger-core/v2/virtual/integration/stateunavailable_test.go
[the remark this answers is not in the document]
ivanshibitov: Ok, I will create a task

ledger-core/v2/virtual/request/vstatereport.go
ivanshibitov: Create task for resolving this issue

PR #113, opened 2020-04-29
Hot fix

MikhailZhukov:
**- What I did**
* Fix hang in slot machine in ctx.AcquireForThisStep(semaphoreReadyToWork)
* Allow text format logging in test

**- How I did it**
@cyraxred remove some optimisation

**- How to verify it**

### DIFF
--- a/ledger-core/v2/virtual/integration/utils/server.go
+++ b/ledger-core/v2/virtual/integration/utils/server.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/insolar/assured-ledger/ledger-core/v2/application/testwalletapi"
 	"github.com/insolar/assured-ledger/ledger-core/v2/configuration"
+	"github.com/insolar/assured-ledger/ledger-core/v2/conveyor/smachine/example"
 	"github.com/insolar/assured-ledger/ledger-core/v2/insolar"
 	"github.com/insolar/assured-ledger/ledger-core/v2/insolar/gen"
 	"github.com/insolar/assured-ledger/ledger-core/v2/insolar/jet"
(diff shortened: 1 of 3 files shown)


ledger-core/v2/virtual/integration/utils/server.go
```diff
@@ -124,7 +125,7 @@ func NewServer(t *testing.T) *Server {
 	virtualDispatcher.Runner = runnerService
 	virtualDispatcher.MessageSender = messageSender
 
-	if err := virtualDispatcher.Init(ctx); err != nil {
+	if err := virtualDispatcher.Init(ctx, &example.MachineLogger{}); err != nil {
```
sexyhamster, line 128: this is wrong, please  copy that structure to integration. We should not use example package

ledger-core/v2/virtual/virtual.go
```diff
@@ -72,10 +72,15 @@ func (lr *Dispatcher) Init(ctx context.Context) error {
 		SlotAliasRegistry: &conveyor.GlobalAliases{},
 	}
 
+	machineConfig := conveyorConfig
```
sexyhamster, line 75: why do you need 
```
conveyorConfig := ...
machineConfig := conveyorConfig
```

MikhailZhukov:
Create new one #114

ledger-core/v2/virtual/virtual.go
cyraxred: to avoid mixing up logs from normal/application SMs and from SMs of the conveyor. SMs of the conveyor only create noise

sexyhamster: My question was - why do we need 2 variables, why not to create machineConfig like it was before?

sexyhamster: oh, I see, sorry it is passed to PulseConveyorConfig, my bad

PR #114, opened 2020-04-29
PLAT-321 short looping fixes

cyraxred:
**- What I did**
- fix for short-loop messing up AcquireForThisStep, added check for sync dependencies after each short-loop
- added exclusion for non-migrate states for short-loop
- fixed unnecessary looping behavior in conveyor

### DIFF
--- a/ledger-core/v2/conveyor/sm_pulse_slot.go
+++ b/ledger-core/v2/conveyor/sm_pulse_slot.go
@@ -212,7 +212,7 @@ func (p *PulseSlotMachine) stepPresentLoop(ctx smachine.ExecutionContext) smachi
 	case !nextPollTime.IsZero():
 		return ctx.WaitAnyUntil(nextPollTime).ThenRepeat()
 	case !p.innerMachine.IsEmpty():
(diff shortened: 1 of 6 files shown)


sexyhamster:
we probably need more tests around this short_looping, and we are not rushing immediate optimizations.

ledger-core/v2/conveyor/sm_pulse_slot.go
```diff
@@ -212,7 +212,7 @@ func (p *PulseSlotMachine) stepPresentLoop(ctx smachine.ExecutionContext) smachi
 	case !nextPollTime.IsZero():
 		return ctx.WaitAnyUntil(nextPollTime).ThenRepeat()
 	case !p.innerMachine.IsEmpty():
```
sexyhamster, line 214: this case could be removed because it duplicates the code after switch

cyraxred: done

sexyhamster: hmm, am I missing something? or you've forgot to push?

PR #116, opened 2020-04-29
PLAT-287: Move to AddInputExt

sexyhamster:
1. switched to AddInputExt in testWalletAPI for guranteed behaviour of testAPI responses
2. moved to context propagation for logging purposes.

### DIFF
--- a/ledger-core/v2/application/testwalletapi/handlers.go
+++ b/ledger-core/v2/application/testwalletapi/handlers.go
@@ -10,11 +10,13 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"sync"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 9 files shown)


ledger-core/v2/application/testwalletapi/handlers.go
```diff
@@ -362,16 +364,42 @@ func (s *TestWalletServer) runWalletRequest(ctx context.Context, req payload.VCa
...
+	ready := &sync.WaitGroup{}
+	ready.Add(1)
+
+	var (
+		fail error
+		res  payload.VCallResult
+	)
+
+	createDefaults := smachine.CreateDefaultValues{
+		Context: ctx,
```
cyraxred, line 379: you don't need to pass ctx here, as you pass it to AddInput

PR #119, opened 2020-04-30
PLAT-299: add test to state exchange

ivanshibitov:
**- What I did**
Add test which check state exchange in case of pulse change 

**- How I did it**

**- How to verify it**


ledger-core/v2/virtual/integration/method_test.go
```diff
@@ -336,3 +336,60 @@ func TestVirtual_Method_WithExecutor(t *testing.T) {
...
+	rawWalletState := makeRawWalletState(t, testBalance)
+	objectRef := reference.NewRecordRef(server.RandomLocalWithPulse())
+	{
+		// send VStateReport: save wallet
+		msg := makeVStateReportEvent(t, objectRef, rawWalletState)
+		require.NoError(t, server.AddInput(ctx, msg))
+	}
+
+	// Change pulse to force send VStateRequest
+	server.IncrementPulse(ctx)
```
sexyhamster, line 386: why do you need to change pulse? Do we already have migrations and stuff?

ivanshibitov: I change pulse since I want to check that we really send VStateRequest to previous executer

PR #120, opened 2020-04-30
PLAT-300: pass whole state in VStateReport

sexyhamster:
- pass whole state in VStateReport
- pass whole objectDescriptor upon modification


ledger-core/v2/virtual/handlers/vstatereport.go
```diff
@@ -45,12 +47,32 @@ func (s *SMVStateReport) Init(ctx smachine.InitializationContext) smachine.State
...
+		immutableExecuteSemaphore smsync.SemaphoreLink
+		mutableExecuteSemaphore   smsync.SemaphoreLink
+
+		deactivated *bool
+	)
+
+	if s.Payload.ProvidedContent != nil && s.Payload.ProvidedContent.LatestDirtyState != nil {
+		objectDescriptor = descriptor.NewObject(
+			objectRef,
+			s.Payload.ProvidedContent.LatestDirtyState.Reference,
```
ivanshibitov, line 67: I would assign 's.Payload.ProvidedContent.LatestDirtyState' to some variable and would use this variable

ledger-core/v2/virtual/handlers/vstatereport.go
```diff
@@ -74,5 +105,8 @@ func (s *SMVStateReport) stepProcess(ctx smachine.ExecutionContext) smachine.Sta
 		panic(throw.NotImplemented())
 	}
 
+	ctx.ApplyAdjustment(immutableExecuteSemaphore.NewDelta(immutablePendingCount))
```
ivanshibitov, line 108: I thought that this should be done inside SMObject

ledger-core/v2/virtual/handlers/vstatereport.go
```diff
@@ -45,12 +47,32 @@ func (s *SMVStateReport) Init(ctx smachine.InitializationContext) smachine.State
...
+	var (
+		objectDescriptor descriptor.Object
+
+		immutableExecuteSemaphore smsync.SemaphoreLink
+		mutableExecuteSemaphore   smsync.SemaphoreLink
+
+		deactivated *bool
+	)
+
+	if s.Payload.ProvidedContent != nil && s.Payload.ProvidedContent.LatestDirtyState != nil {
```
ivanshibitov, line 64: In what case do they equal to 'nil'?

ledger-core/v2/virtual/execute/execute.go
```diff
@@ -285,15 +287,41 @@ func (s *SMExecute) stepSaveNewObject(ctx smachine.ExecutionContext) smachine.St
 	case insolar.RequestSideEffectActivate:
 		_, prototype, memory = executionNewState.Activate()
 		action = func(state *object.SharedState) {
-			state.Info.SetDescriptor(&prototype, memory)
+			stateID := NewStateID(s.pulseSlot.PulseData().GetPulseNumber(), memory)
```
ivanshibitov, line 290: It looks the same like in 'case insolar.RequestSideEffectAmend'

ledger-core/v2/virtual/handlers/vstatereport.go
sexyhamster: if it was not requested via VStateRequest

ledger-core/v2/virtual/handlers/vstatereport.go
ruz: this is SO wrong :) counters that are in VStateReport shouldn't change semaphores, they should change counters.

sexyhamster: you are right, when I've started the ticket there were no counters so I assumed wrong behavior. Fixed

bigbes:
👌

PR #127, opened 2020-05-02
Cleanup of ref marshaling 

cyraxred:
**- What I did**
- Remove jet leftovers from reference and other code
- Moved marshal/unmarshal logic to package methods to ensure immutable nature of references
- Clarified deprecations of reference.Local/Global methods


ledger-core/v2/reference/marshal_local.go
```diff
@@ -0,0 +1,133 @@
...
+		v = decoded.GetLocal()
+	case nil:
+		//
+	default:
+		err = throw.W(err, "unexpected type %T when unmarshal reference.Local", repr)
+	}
+	return
+}
+
+func MarshalLocalJSON(v LocalHolder) (b []byte, err error) {
```
AndreyBronin, line 91: The method actually marshall ref to string form

cyraxred: well, it mimics MarshalJSON signature for simpler use, hence there are []byte here

ledger-core/v2/insolar/payload/payload.go
```diff
@@ -15,15 +15,12 @@ import (
 type Type uint32
 
 //go:generate stringer -type=Type
-// //go:generate protoc -I=. -I=$GOPATH/src --gogoslick_out=./ payload.proto
+// xgo:generate protoc -I=. -I=$GOPATH/src --gogoslick_out=./ payload.proto
```
ruz, line 18: xgo?

PR #128, opened 2020-05-02
PLAT-82 record-message serializer

cyraxred:
**- What I did**
* protoc-based code generation that does all necessary work, including:
** no more nullable=false etc
** zero-size types are skipped without nullable (zero-able, not null-able)
** types are registered for deserialization
** code and checks for notation compliance
** generation of head(s)
** etc
* protoc-gen-ins with extra options and logic
* some tricks to change protoc-gen-gogo behavior without fork


ledger-core/v2/reference/global_test.go
```diff
@@ -42,17 +42,65 @@ func randLocal() Local {
 	return id
 }
 
+func flipHeader(b []byte) []byte {
+	// Encoder uses BigEndian, while AsBytes uses LittleEndian for the header
+	// b[0], b[1], b[2], b[3] = b[3], b[2], b[1], b[0]
+	return b
```
sexyhamster, line 48: something is missing over here or function is not needed anymore

cyraxred: This is stashed as first I'd like to merge this branch, and then fix the encoding.

ledger-core/v2/reference/scope.go
```diff
@@ -27,47 +31,79 @@ const ( // super-scopes
 )
 
 const ( // super-scopes
-	LifelineSuperScope    Scope = 0x04 * baseScopeLifeline
-	LocalDomainSuperScope Scope = 0x04 * baseScopeLocalDomain
-	GlobalSuperScope      Scope = 0x04 * baseScopeGlobal
+	superScopeLifeline    Scope = 0x04 * baseScopeLifeline
```
ivanshibitov, line 34: I would give name to 0x04

cyraxred: done
593407f8c87af0ebb5d4301e5e2983510f2aad7d

PR #137, opened 2020-05-08
PLAT-278: Convert struct object.Catalog to interface

MikhailZhukov:
**- What I did**
Simplify testing SMExecute

**- How I did it**
Convert struct `object.Catalog` to interface and use it as dependency in state machine. This allow mock interface of `object.Catalog` and access to internal of state machine object

**- How to verify it**
Run test


ledger-core/v2/virtual/integration/utils/server.go
```diff
@@ -119,6 +122,9 @@ func NewServer(t *testing.T) *Server {
 	if err := virtualDispatcher.Init(ctx); err != nil {
 		panic(err)
 	}
+	//
+	// objectCatalog = object.NewLocalCatalog()
+	// virtualDispatcher.Conveyor.AddDependency(objectCatalog)
```
ruz, line 127: please remove commented out code

ledger-core/v2/conveyor/conveyor.go
```diff
@@ -107,6 +107,14 @@ func (p *PulseConveyor) AddDependency(v interface{}) {
 	p.slotMachine.AddDependency(v)
 }
 
+func (p *PulseConveyor) AddDependencyByName(id string, v interface{}) {
+	p.slotMachine.AddDependencyByName(id, v)
+}
```
ruz, line 112: this is not in use

ledger-core/v2/conveyor/conveyor.go
```diff
@@ -107,6 +107,14 @@ func (p *PulseConveyor) AddDependency(v interface{}) {
...
+
+func (p *PulseConveyor) AddDependencyByType(vt interface{}, v interface{}) {
```
ruz, line 114: from what I see this is not required or at least wrongly named. In every use you pass pointer to a value as first argument and value as second argument. Meaning of this is not clear to me.

ledger-core/v2/virtual/integration/utils/server.go
MikhailZhukov: Fix it

ledger-core/v2/conveyor/conveyor.go
MikhailZhukov: Remove this code

ledger-core/v2/conveyor/conveyor.go
MikhailZhukov: When we use SlotMachine.AddDependency with interface go type system return name of struct but not an interface. But when DI try inject interface to state machine it can't find name (name of struct != name of interface). @cyraxred say what need to create new function for that case. I agree that name of function must be better but i don't know how to name it

ledger-core/v2/virtual/integration/utils/server.go
```diff
@@ -67,6 +69,7 @@ func NewServer(t *testing.T) *Server {
 	var (
 		PulseManager *pulsemanager.PulseManager
 		Pulses       *pulse.StorageMem
+		// objectCatalog object.Catalog
```
ruz, line 72: commented out code

ledger-core/v2/conveyor/conveyor.go
```diff
@@ -107,6 +107,10 @@ func (p *PulseConveyor) AddDependency(v interface{}) {
 	p.slotMachine.AddDependency(v)
 }
 
+func (p *PulseConveyor) AddDependencyByVar(v interface{}) {
```
ruz, line 110: I don't understand picked name, for example in the following code name doesn't explain difference between two calls to different variants of AddDependency:  
```
	slotMachine.AddDependency(&pulseSlot)

	var objectCatalog object.Catalog = object.NewLocalCatalog()
	slotMachine.AddDependencyByVar(&objectCatalog)
```

bigbes:
https://github.com/insolar/assured-ledger/pull/145

ledger-core/v2/virtual/virtual.go
```diff
@@ -93,6 +95,11 @@ func (lr *Dispatcher) Init(ctx context.Context) error {
 	lr.Conveyor.AddDependency(lr.runnerAdapter)
 	lr.Conveyor.AddDependency(lr.messageSenderAdapter)
 
+	if lr.ObjectCatalog == nil {
```
ivanshibitov, line 98: In what case won't It be equal to nil?

MikhailZhukov: It can be not nil if it will set before Dispatcher.Init

PR #144, opened 2020-05-08
NOISSUE: implement timeouts for API

sexyhamster:
**- What I did**
I've added timeout for all API integration tests, and added timeout for one of tests that could freeze in case of panic

### DIFF
--- a/ledger-core/v2/virtual/integration/constructor_test.go
+++ b/ledger-core/v2/virtual/integration/constructor_test.go
@@ -196,7 +196,7 @@ func TestVirtual_Constructor_WithExecutor(t *testing.T) {
 		server.SendMessage(ctx, msg)
 
 		select {
-		case <-time.After(3 * time.Second):
+		case <-time.After(10 * time.Second):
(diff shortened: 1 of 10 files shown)


ledger-core/v2/virtual/integration/constructor_test.go
```diff
@@ -194,6 +195,10 @@ func TestVirtual_Constructor_WithExecutor(t *testing.T) {
 
 		server.SendMessage(ctx, msg)
 
-		<-testIsDone
+		select {
+		case <-time.After(3 * time.Second):
```
ruz, line 199: 3 seconds is often way to low for CI environments

sexyhamster: It is set on other tests to 3 seconds and was not triggering fails for now. I'm not sure whether it is better to have it in single constant or could be set individually for each test. If tests will start flapping we will adjust timeouts

sexyhamster: updated timeout to be 10 seconds.

PR #147, opened 2020-05-12
PLAT-279:  process finish execution

ivanshibitov:
**- What I did**
1. Add Migration for SmExecute
2. Send VDelegatedFinished if pulse was changed during execution

**- How I did it**

**- How to verify it**


ledger-core/v2/virtual/execute/execute.go
```diff
@@ -415,6 +423,34 @@ func (s *SMExecute) stepSaveNewObject(ctx smachine.ExecutionContext) smachine.St
...
+		ResultFlags:        nil,
+		CallOutgoing:       s.Payload.CallOutgoing,
+		CallIncoming:       reference.Local{},
+		EntryHeadHash:      nil,
+		DelegationSpec:     nil,
+		DelegatorSignature: nil,
+		ObjectBody:         s.executionNewState.Result.Memory,
+	}
+
+	s.pulseSlot.PulseData().NextPulseNumber()
```
bigbes: ??

ledger-core/v2/virtual/execute/execute.go
```diff
@@ -440,6 +474,9 @@ func (s *SMExecute) setNewState(prototype reference.Global, memory []byte) func(
 			memory,
 			parentReference,
 		))
+		if !s.migrationHappened {
```
bigbes, line 477: Why we shouldn't decrement here?

ivanshibitov: We just don't need them anymore, since pulse was changed

bigbes: Then it's just redundant if, that may lead to confusion, imho

ledger-core/v2/virtual/execute/execute.go
```diff
@@ -100,6 +102,7 @@ func (s *SMExecute) prepareExecution(ctx smachine.InitializationContext) {
 }
 
 func (s *SMExecute) Init(ctx smachine.InitializationContext) smachine.StateUpdate {
+	ctx.SetDefaultMigration(s.migrateDuringExecution)
```
bigbes, line 105: Not sure if it's right place.
If we don't start an execution - then we don't need to set "migrationHappened" and we need to stop executing (not sure about executing part, more like "increment counter part")

ledger-core/v2/virtual/integration/delegated_finished_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	"github.com/insolar/assured-ledger/ledger-core/v2/insolar/gen"
+	"github.com/insolar/assured-ledger/ledger-core/v2/insolar/payload"
+	"github.com/insolar/assured-ledger/ledger-core/v2/instrumentation/inslogger"
+	"github.com/insolar/assured-ledger/ledger-core/v2/virtual/integration/utils"
+)
+
+// 1. Send CallRequest
+// 2. Since there is no state SM send VSateRequest
+// 3. We catch VSateRequest. Change pulse. Send VStateReport
```
bigbes, line 23: Not sure if it's right place to change pulse here.
I think it should be changed inside contract execution

ledger-core/v2/virtual/execute/execute.go
ivanshibitov: Possible, yes, It's better to decrement in any case

ledger-core/v2/virtual/execute/execute.go
ivanshibitov: Yes, move it to stepUpdatePendingCounters step

ledger-core/v2/virtual/integration/delegated_finished_test.go
ivanshibitov: mock executor and change pulse inside of this mock

ledger-core/v2/virtual/execute/execute.go
```diff
@@ -441,6 +475,8 @@ func (s *SMExecute) setNewState(prototype reference.Global, memory []byte) func(
 			memory,
 			parentReference,
 		))
+
+		state.DecrementPotentialPendingCounter(!s.execution.Unordered)
```
ruz, line 479: Bugs here:
- setNewState action on shared data is not always called, for example it is not called when there is no new state produced, this should be covered with a unit test on stepSaveNewObject
- counters of potential pendings shouldn't be changed after pulse change, they are not potential anymore. As we don't use other measures to finalize object state, implemented behaviour will result in different VStateReports on multiple VStateRequests.

ledger-core/v2/virtual/execute/execute.go
ruz: as I wrote in other comment, decrementing in either case result in a bug

ledger-core/v2/virtual/execute/execute.go
```diff
@@ -416,6 +424,32 @@ func (s *SMExecute) stepSaveNewObject(ctx smachine.ExecutionContext) smachine.St
...
+		CallType:           s.Payload.CallType,
+		CallFlags:          s.Payload.CallFlags,
+		Callee:             s.execution.Object,
+		ResultFlags:        nil,
+		CallOutgoing:       s.Payload.CallOutgoing,
+		CallIncoming:       reference.Local{},
+		EntryHeadHash:      nil,
+		DelegationSpec:     nil,
+		DelegatorSignature: nil,
+		ObjectBody:         s.executionNewState.Result.Memory,
```
ruz, line 445: Why do we send it unconditionally? I believe we should send new data only when a new state is produced. I'm sure that object's memory is not enough. At least it should contain identifier of the new state. More should be learned from VStateReport.

ledger-core/v2/virtual/execute/execute.go
ivanshibitov: Right now, don't change counters if pulse was changed.
In the future, when VStateReport get state from immutable SM, we will just remove this 'if' statement

ledger-core/v2/virtual/execute/execute.go
ivanshibitov: 1. add check if execution is unordered - don't send object body 
2. add ObjectState to VDelegatedRequestFinished

ledger-core/v2/virtual/execute/execute.go
ivanshibitov: 1. Don't change counters after pulse chafe
2. fix problem in case of there is no new state

ledger-core/v2/virtual/execute/execute.go
```diff
@@ -441,6 +492,8 @@ func (s *SMExecute) setNewState(prototype reference.Global, memory []byte) func(
 			memory,
 			parentReference,
 		))
+
+		s.updateCounters()
```
bigbes, line 496: that won't work

ledger-core/v2/virtual/execute/execute.go
```diff
@@ -402,23 +414,62 @@ func (s *SMExecute) stepSaveNewObject(ctx smachine.ExecutionContext) smachine.St
...
+
+	if s.migrationHappened {
+		return ctx.Jump(s.stepSendDelegatedRequestFinished)
+	}
+
+	return ctx.Jump(s.stepSendCallResult)
+}
+
+func (s *SMExecute) stepSendDelegatedRequestFinished(ctx smachine.ExecutionContext) smachine.StateUpdate {
+
```
bigbes, line 437: excessive newline

ledger-core/v2/virtual/execute/execute_test.go
```diff
@@ -79,6 +79,8 @@ func TestSMExecute_IncreasePendingCounter(t *testing.T) {
 		JumpMock.Set(CheckWrapper(stepChecker, t)).
 		UseSharedMock.Set(CallSharedDataAccessor)
```
bigbes, line 81: you may merge two that blocks

ledger-core/v2/virtual/execute/execute.go
ivanshibitov: In next pr

ledger-core/v2/virtual/execute/execute_test.go
ivanshibitov: fixing in next pr

PR #149, opened 2020-05-13
PLAT-277: Await pending ordered SM

dgovorukhin:
**- What I did**
Added AwaitOrderedPendingSM and acquire SMObject.MutableExecute if we have more than zero in SMObject.ActiveMutablePendingCount. It needs to stop ordered operation on new execute before old execute will complete ordered operations.
**- How I did it**
Created a new AwaitOrderedPendingSM for acquiring SMObject.MutableExecute. we should release this semaphore on VDelegatedRequestFinished. (shared BargIn in SMObject  for stop AwaitOrderedPendingSM )
**- How to verify it**

https://insolar.atlassian.net/browse/PLAT-277


ledger-core/v2/virtual/object/object.go
```diff
@@ -39,6 +39,8 @@ type Info struct {
 	MutableExecute   smachine.SyncLink
 	ReadyToWork      smachine.SyncLink
 
+	AwaitPendingOrdered smachine.BargeIn
```
ivanshibitov, line 42: Who has to fill this field?

ledger-core/v2/virtual/object/orderedPending.go
```diff
@@ -0,0 +1,37 @@
...
+func (sm *AwaitOrderedPendingSM) Init(ctx smachine.InitializationContext) smachine.StateUpdate {
+	if !ctx.Acquire(sm.sync) {
+		panic("failed to acquire semaphore")
+	}
+
+	// Sync will be released on machine stop.
+	sm.stop = ctx.NewBargeIn().WithStop()
+
+	return ctx.Jump(func(ctx smachine.ExecutionContext) smachine.StateUpdate {
+		return ctx.Sleep().ThenRepeat()
```
ruz, line 35: this looks wrong. if SM would be woken up for any reason then it will either panic with failure to acquire lock or acquires twice or may be MutableExecute checks double acquire, it's not clear. Create a separate step for spin sleep.

ledger-core/v2/virtual/object/object.go
```diff
@@ -194,3 +199,16 @@ func (sm *SMObject) stepReadyToWork(ctx smachine.ExecutionContext) smachine.Stat
...
+	syncSM := AwaitOrderedPendingSM{
+		sync: sm.MutableExecute,
+	}
+
+	// syncSM acquire MutableExecute semaphore in init step.
+	ctx.InitChild(func(ctx smachine.ConstructionContext) smachine.StateMachine {
+		return &syncSM
+	})
+
+	sm.AwaitPendingOrdered = syncSM.stop
```
ruz, line 213: @cyraxred, is it allowed to copy barge in like this from other state machine?

dgovorukhin: @ruz As I can see in player.go BargIn copied to shared data like this
Screenshot 2020-05-14 at 11 54 03
I can change to a pointer if you not like this copy)

ledger-core/v2/virtual/object/orderedPending.go
dgovorukhin: @ruz Fixed

ledger-core/v2/virtual/object/object.go
ruz: it's different. In the code actually to cases: [case 1](https://github.com/insolar/assured-ledger/blob/fdad6296dafbb8f926036555a0d790023bac2fa1/ledger-core/v2/conveyor/smachine/example/player.go#L112) and [case 2](https://github.com/insolar/assured-ledger/blob/fdad6296dafbb8f926036555a0d790023bac2fa1/ledger-core/v2/conveyor/smachine/example/player.go#L135). In both cases bargeins are injected into shared data. In case one it's access to shared data. In case two it's publish of a new shared data.

dgovorukhin: Got it.

dgovorukhin: @ruz Fixed

ledger-core/v2/virtual/object/object.go
```diff
@@ -202,3 +207,16 @@ func (sm *SMObject) stepReadyToWork(ctx smachine.ExecutionContext) smachine.Stat
...
+func (sm *SMObject) createWaitPendingOrderedSM(ctx smachine.ExecutionContext) {
+	syncSM := AwaitOrderedPendingSM{
+		sync: sm.MutableExecute,
+	}
+
+	// syncSM acquire MutableExecute semaphore in init step.
+	ctx.InitChildWithPostInit(func(ctx smachine.ConstructionContext) smachine.StateMachine {
+		return &syncSM
+	}, func() {
+		sm.AwaitPendingOrdered = &syncSM.stop
```
cyraxred, line 220: BargeIn is by-value, so no need for ptr here

ledger-core/v2/virtual/object/object.go
cyraxred: @ruz actually, bargein, like synclink is by-value, so no need for ptr

ledger-core/v2/virtual/object/object.go
dgovorukhin: Fixed

PR #152, opened 2020-05-13
PLAT-282: Send VStateReport if pulse was changed

MikhailZhukov:
**- What I did**
Send VStateReport when pulse was changed

**- How I did it**
Add migration to SMObject which send message VStateReport

**- How to verify it**
Run test


ledger-core/v2/virtual/object/object.go
```diff
@@ -194,3 +197,48 @@ func (sm *SMObject) stepReadyToWork(ctx smachine.ExecutionContext) smachine.Stat
...
+		err := svc.SendRole(goCtx, &msg, insolar.DynamicRoleVirtualExecutor, sm.Reference, pulseNumber)
+		if err != nil {
+			panic(err)
+		}
+	}).Send()
+	return ctx.Jump(sm.migrateTransition)
+}
+
+func (sm *SMObject) migrateDuringExecution(ctx smachine.MigrationContext) smachine.StateUpdate {
+	ctx.Log().Warn("Migration happened: " + sm.pulseSlot.CurrentPulseNumber().String())
```
ivanshibitov, line 241: Do you really need it?

ledger-core/v2/virtual/object/object.go
```diff
@@ -202,3 +205,38 @@ func (sm *SMObject) stepReadyToWork(ctx smachine.ExecutionContext) smachine.Stat
...
+
+func (sm *SMObject) stepSendVStateReport(ctx smachine.ExecutionContext) smachine.StateUpdate {
+	var (
+		pulseNumber = sm.pulseSlot.CurrentPulseNumber()
+	)
+
+	objDescriptor := sm.Descriptor()
+	prototype, _ := objDescriptor.Prototype()
+	msg := payload.VStateReport{
+		ImmutablePendingCount: int32(sm.ActiveImmutablePendingCount),
```
ruz, line 217: this is incorrect, it should be sum of active and potential

MikhailZhukov: Fix pending counting

ledger-core/v2/network/messagesender/adapter/init.go
```diff
@@ -12,12 +12,19 @@ import (
...
 
-type MessageSender struct {
+//go:generate minimock -i github.com/insolar/assured-ledger/ledger-core/v2/network/messagesender/adapter.MessageSender -o ./ -s _mock.go -g
+type MessageSender interface {
+	PrepareSync(ctx smachine.ExecutionContext, fn func(svc messagesender.Service)) smachine.SyncCallRequester
+	PrepareAsync(ctx smachine.ExecutionContext, fn func(svc messagesender.Service) smachine.AsyncResultFunc) smachine.AsyncCallRequester
+	PrepareNotify(ctx smachine.ExecutionContext, fn func(svc messagesender.Service)) smachine.NotifyRequester
+}
+
+type ParallelMessageSender struct {
```
bigbes, line 22: Why parallel?

ledger-core/v2/virtual/execute/update_pending_counters_test.go
```diff
@@ -225,8 +225,8 @@ func Test_SlotMachine_Increment_Pending_Counters(t *testing.T) {
 		QueryRoleMock.Return([]reference.Global{gen.Reference()}, nil)
 	pulses := pulse.NewStorageMem()
 	messageSender := messagesender.NewDefaultService(publisherMock, jetCoordinatorMock, pulses)
-	messageSenderAdapter := messagesenderadapter.CreateMessageSendService(ctx, messageSender)
-	slotMachine.AddDependency(messageSenderAdapter)
+	var messageSenderAdapter messagesenderadapter.MessageSender = messagesenderadapter.CreateMessageSendService(ctx, messageSender)
```
bigbes, line 228: use, instead:
```suggestion
	messageSenderAdapter := messagesenderadapter.CreateMessageSendService(ctx, messageSender)
```

ledger-core/v2/virtual/integration/vstatereport_if_pulse_changed_test.go
```diff
@@ -0,0 +1,92 @@
+// Copyright 2020 Insolar Network Ltd.
```
bigbes, line 1: Rename that file to simple `vstatereport_test.go`, i don't think that 1000 files (1 file  with `_test.go` suffix for each testcase) is a potentially good idea

ledger-core/v2/virtual/execute/update_pending_counters_test.go
ruz: this is is intentional. CreateMessageSendService returns exact implementation. We need interface.

ledger-core/v2/network/messagesender/adapter/init.go
MikhailZhukov: Because we have only one implementation and i try to decide witch name better shows it destination. This implementation use `smachine.StartChannelWorkerParallelCalls` and i decide that Parallel is important property of this implementation

ledger-core/v2/virtual/execute/update_pending_counters_test.go
MikhailZhukov: Yes, it is very important use interface here

ledger-core/v2/virtual/integration/vstatereport_if_pulse_changed_test.go
MikhailZhukov: I think what we need better name for this test because we have vstatereport in handlers too. Maybe name must show what this message show SMObject

PR #153, opened 2020-05-13
PLAT-327: proxy tolerance and isolation flags

sexyhamster:
Added proxy/outogoing calls propagation for tolerance and state flag


ledger-core/v2/application/preprocessor/templates/proxy.go.tpl
```diff
@@ -180,14 +183,22 @@ func (r *{{ $.ContractType }}) {{ $method.Name }}{{if $method.Immutable}}AsMutab
 		return {{ $method.ResultsWithErr }}
 	}
 
+	{{ if $method.Immutable }}
```
ruz, line 186: you don't need this `if`. Every method has two variants. This one is 'Mutable'. If method `Xxx` is marked by author as immutable then this block generates method `XxxAsMutable`, so it's still call with `CallTolerable` and `CallDirty `.

sexyhamster: fixed

ledger-core/v2/application/preprocessor/templates/proxy.go.tpl
```diff
@@ -20,6 +20,7 @@
 package {{ .PackageName }}
 
 import (
+	"github.com/insolar/assured-ledger/ledger-core/v2/insolar/payload"
```
ruz, line 23: we do all imports [in the code](https://github.com/insolar/assured-ledger/blob/08a8da167bc74a667002ec639ed59c9b4e105030/ledger-core/v2/application/preprocessor/main.go#L609)

sexyhamster: thats not entirely true, https://github.com/insolar/assured-ledger/blob/08a8da167bc74a667002ec639ed59c9b4e105030/ledger-core/v2/application/preprocessor/templates/initialization.go.tpl#L23
I didn't want to touch this generateImport function, because off poor design with wrapper/notwrapper signature. I'll think what could be done, looks ugly anyway

ruz: poor design is a poor excuse. Consistency wins while we're looking for beautiful looking way.

sexyhamster: ok, now all imports are in generator, nothing lives in templates to avoid confusions in futute

ledger-core/v2/application/builtin/proxy/testwallet/testwallet.go
```diff
@@ -110,7 +111,8 @@ func (r *Wallet) GetPrototype() (reference.Global, error) {
 		var ret1 *foundation.Error
 		ret[1] = &ret1
 
-		res, err := common.CurrentProxyCtx.CallMethod(r.Reference, false, false, "GetPrototype", make([]byte, 0), PrototypeReference)
+		res, err := common.CurrentProxyCtx.CallMethod(
```
ivanshibitov, line 114: looks like you do it accidentally =)

sexyhamster: nope, the line width exceeded screen, so I've made new line here

ivanshibitov: ok. it looks with additional newline. my bad

PR #154, opened 2020-05-13
PLAT-198: logger fix for SendNotify logging propagation

sexyhamster:
subj

### DIFF
--- a/ledger-core/v2/network/messagesender/adapter/init.go
+++ b/ledger-core/v2/network/messagesender/adapter/init.go
@@ -38,10 +38,10 @@ func (a *MessageSender) PrepareAsync(
 
 func (a *MessageSender) PrepareNotify(
 	ctx smachine.ExecutionContext,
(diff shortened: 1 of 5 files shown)


ledger-core/v2/network/messagesender/adapter/init.go
```diff
@@ -38,10 +38,10 @@ func (a *MessageSender) PrepareAsync(
 
 func (a *MessageSender) PrepareNotify(
 	ctx smachine.ExecutionContext,
-	fn func(svc messagesender.Service),
+	fn func(logger smachine.Logger, svc messagesender.Service),
 ) smachine.NotifyRequester {
 	return a.exec.PrepareNotify(ctx, func(interface{}) {
-		fn(a.svc)
+		fn(ctx.LogAsync(), a.svc)
```
ruz, line 44: shouldn't we provide some sort of AdapterContext with a few methods?

PR #155, opened 2020-05-13
NOISSUE: rewrite runner adapter to use less goroutines

bigbes:
previous worker algo was:
* SM sleeps
* spawn goroutine that will prepare execution
* spawn goroutine that will execute request
* wait until it finished or outgoing request
* notify SM that called adapter

new worker is:
* SM sleeps
* spawn goroutine that will execute request
* eventually it will wakeup SM that called adapter (when
outgoing/request finished)

It can be implemented using `DelegateAndSendResult` function on
AdapterCall that is spawned by ExecutionAdapter that allows us
to wakeup parent SM (using fn return argument)

Also, when function is called, worker catches request, intercepts it
(using runnerServiceInterceptor), fills internal state and calls the
right function of real runner service.

Previously runner could work without adapter, but now it 'kinda' depends
on adapter (however it can work without it)

Also applied a couple of renames


ledger-core/v2/runner/executor/builtin/builtin.go
```diff
@@ -49,11 +50,11 @@ func New(stub common.RunnerRPCStub) *Runner {
 
 func (r *Runner) CallConstructor(
 	_ context.Context,
-	callCtx *insolar.LogicCallContext,
+	callCtx *callcontext.LogicCallContext,
```
cyraxred, line 53: may be call.Context or call.LogicContext ?

ledger-core/v2/runner/executor/builtin/builtin.go
```diff
@@ -49,11 +50,11 @@ func New(stub common.RunnerRPCStub) *Runner {
...
 	codeRef reference.Global,
 	name string,
-	args insolar.Arguments,
-) ([]byte, insolar.Arguments, error) {
+	args []byte,
```
cyraxred, line 56: Please consider use of ByteString  for immutable bytes

ledger-core/v2/application/cmd/insgocc/builder.go
```diff
@@ -72,7 +73,7 @@ func (cb *contractsBuilder) clean() {
 }
 
 func (cb *contractsBuilder) parseContract(name string) (*preprocessor.ParsedFile, error) {
-	return preprocessor.ParseFile(cb.getContractPath(name), insolar.MachineTypeGoPlugin)
+	return preprocessor.ParseFile(cb.getContractPath(name), machinetype.Builtin)
```
cyraxred, line 76: I'd rather name it machine.Type and machine.Builtin

ledger-core/v2/runner/machinetype/init.go
```diff
@@ -0,0 +1,20 @@
+// Copyright 2020 Insolar Network Ltd.
+// All rights reserved.
+// This material is licensed under the Insolar License version 1.0,
+// available at https://github.com/insolar/assured-ledger/blob/master/LICENSE.md.
+
+package machinetype
```
cyraxred, line 6: 1. there is no initialization in this file - please rename it to machine.go
2. I'd suggest to rename package to just "machine"

ledger-core/v2/runner/machinetype/init.go
```diff
@@ -0,0 +1,20 @@
...
+
+// MachineType is a type of virtual machine
+type MachineType int
```
cyraxred, line 9: rename to just Type

ledger-core/v2/runner/machinetype/init.go
```diff
@@ -0,0 +1,20 @@
...
+
+// Real constants of MachineType
+const (
+	Unknown MachineType = iota
+	Builtin
+	LastID
+)
+
+func (m MachineType) Equal(other MachineType) bool {
```
cyraxred, line 18: If this is for proto, then please use "casttype" option instead of "customtype" for fields with MachineType. And delete this method

ledger-core/v2/runner/manager/manager.go
```diff
@@ -0,0 +1,32 @@
...
+	"context"
+
+	"github.com/insolar/assured-ledger/ledger-core/v2/reference"
+	"github.com/insolar/assured-ledger/ledger-core/v2/runner/callcontext"
+)
+
+//go:generate minimock -i github.com/insolar/assured-ledger/ledger-core/v2/runner/manager.MachineLogicExecutor -o ./ -s _mock.go -g
+
+// MachineLogicExecutor is an interface for implementers of one particular machine type
+type MachineLogicExecutor interface {
```
cyraxred, line 18: may be change it to machine.Executor?

ledger-core/v2/runner/runner.go
```diff
@@ -8,43 +8,41 @@ package runner
...
 	"github.com/insolar/assured-ledger/ledger-core/v2/runner/requestresult"
 	"github.com/insolar/assured-ledger/ledger-core/v2/vanilla/throw"
 	"github.com/insolar/assured-ledger/ledger-core/v2/virtual/descriptor"
 )
 
 type Service interface {
-	ExecutionStart(ctx context.Context, execution execution.Context) (*executionupdate.ContractExecutionStateUpdate, uuid.UUID, error)
-	ExecutionContinue(ctx context.Context, id uuid.UUID, result []byte) (*executionupdate.ContractExecutionStateUpdate, error)
-	ExecutionAbort(ctx context.Context, id uuid.UUID)
-	ExecutionClassify(ctx context.Context, execution execution.Context) calltype.ContractCallType
```
cyraxred, line 34: Classify will be needed

ledger-core/v2/runner/runner.go
```diff
@@ -56,58 +54,54 @@ func (r *DefaultService) stopExecution(id uuid.UUID) error { // nolint
 	return nil
 }
 
-func (r *DefaultService) getExecutionSink(id uuid.UUID) *executionEventSink {
+func (r *DefaultService) getExecutionSink(id callcontext.CorrelationID) *executionEventSink {
```
cyraxred, line 57: call.ID or call.CorrelationID?

ledger-core/v2/virtual/execute/execute.go
```diff
@@ -351,23 +338,9 @@ func (s *SMExecute) stepExecuteOutgoing(ctx smachine.ExecutionContext) smachine.
...
-		defer statemachine.LogAsyncTime(asyncLogger, time.Now(), "ExecutionStart")
-
-		newState, err := svc.ExecutionContinue(goCtx, executionID, outgoingResult)
-
-		return func(ctx smachine.AsyncResultContext) {
-			s.executionNewState = newState
-			s.executionError = err
-		}
-	}).DelayedStart().Sleep().ThenJump(s.stepExecuteDecideNextStep)
+	return s.runner.PrepareExecutionContinue(ctx, s.run, outgoingResult, func() {}).DelayedStart().Sleep().ThenJump(s.stepExecuteDecideNextStep)
```
cyraxred, line 343: may be nil instead of func(){}?

ledger-core/v2/runner/runner.go
bigbes: @sexyhamster will reintroduce it in his patch

ledger-core/v2/runner/adapter.go
```diff
@@ -0,0 +1,144 @@
...
+	worker *worker
+}
+
+func (a *ServiceAdapter) PrepareExecutionStart(ctx smachine.ExecutionContext, execution execution.Context, fn func(*RunState)) smachine.AsyncCallRequester {
+	if fn == nil {
+		panic(throw.IllegalValue())
+	}
+
+	return a.exec.PrepareAsync(ctx, func(arg interface{}) smachine.AsyncResultFunc {
+		state := arg.(Service).ExecutionStart(execution)
```
sexyhamster, line 97: why not a.svc?

ledger-core/v2/runner/runner.go
sexyhamster: as we've discussed, please put it back :)

bigbes: In next PR, if you don't mind.

ledger-core/v2/runner/eventsink.go
```diff
@@ -6,53 +6,47 @@
...
 	"github.com/insolar/assured-ledger/ledger-core/v2/runner/execution"
 	descriptor "github.com/insolar/assured-ledger/ledger-core/v2/runner/executionevent"
 	"github.com/insolar/assured-ledger/ledger-core/v2/runner/executionupdate"
 	"github.com/insolar/assured-ledger/ledger-core/v2/runner/requestresult"
 )
 
 type executionEventSink struct {
+	id      call.ID
 	context execution.Context
 	output  chan *executionupdate.ContractExecutionStateUpdate
```
cyraxred, line 19: You don't need to use channel output - a direct pointer to RunState will be fine. There is no racing here, as RunState will only be accessed after triggering a callback.
Same for input - this is only one time assign and one time read.

ledger-core/v2/runner/runner.go
cyraxred: I'm ok

ledger-core/v2/runner/eventsink.go
bigbes: Channel for input (outgoing result) is needed to wakeup goroutine that executes contract

bigbes:
rebased onto latest master

PR #156, opened 2020-05-14
PLAT-345: switch to PrepareAsync because we need context with logger

sexyhamster:
We should not use context in detached calls Notify/Async


ledger-core/v2/virtual/object/object.go
```diff
@@ -184,9 +184,14 @@ func (sm *SMObject) stepGetObjectState(ctx smachine.ExecutionContext) smachine.S
...
-		_ = svc.SendRole(goCtx, &msg, insolar.DynamicRoleVirtualExecutor, ref, prevPulse)
-	}).Send()
+	sm.messageSender.PrepareAsync(ctx, func(svc messagesender.Service) smachine.AsyncResultFunc {
+		err := svc.SendRole(goCtx, &msg, insolar.DynamicRoleVirtualExecutor, ref, prevPulse)
+		return func(ctx smachine.AsyncResultContext) {
+			if err != nil {
+				ctx.Log().Error("failed to send state request", err)
+			}
+		}
+	}).Start()
```
bigbes, line 194: WithFlags here?

ledger-core/v2/virtual/execute/execute.go
```diff
@@ -334,12 +334,14 @@ func (s *SMExecute) stepExecuteOutgoing(ctx smachine.ExecutionContext) smachine.
...
 			err := svc.SendRole(goCtx, msg, insolar.DynamicRoleVirtualExecutor, object, pulseNumber)
-			if err != nil {
-				panic(err)
+			return func(ctx smachine.AsyncResultContext) {
+				if err != nil {
+					ctx.Log().Error("failed to send message", err)
+				}
 			}
-		}).DelayedSend().Sleep().ThenJump(s.stepExecuteContinue) // we'll wait for barge-in WakeUp here, not adapter
+		}).WithFlags(0).DelayedStart().Sleep().ThenJump(s.stepExecuteContinue) // we'll wait for barge-in WakeUp here, not adapter
```
bigbes, line 344: Let's split it to 2 different statements?

PR #157, opened 2020-05-14
PLAT-294 stepGetObjectState with delay

aalekseev-ru:
**- What I did**
add waiting before sending VStateReport
**- How I did it**
InitReasonCTMethod now jumps into stepWaitState
then if 10% of pulse time is done and object is not ready we send VStateReport once
**- How to verify it**
run tests


ledger-core/v2/virtual/object/object.go
```diff
@@ -152,12 +157,16 @@ func (sm *SMObject) Init(ctx smachine.InitializationContext) smachine.StateUpdat
 		return ctx.Stop()
 	}
 
+	waitDuration := time.Second * time.Duration(sm.pulseSlot.PulseData().NextPulseDelta) / 10
```
ivanshibitov, line 160: May be move 10% to config?

aalekseev-ru: @ivanshibitov which one? I knew only configuration.go, but it's more about components, not constants. I can move that to object SM constant

bigbes: Just move it to constant, for now

ledger-core/v2/virtual/integration/object_test.go
```diff
@@ -57,7 +57,13 @@ func TestInitViaCTMethod(t *testing.T) {
 
 	select {
 	case <-requestIsDone:
-	case <-time.After(10 * time.Second):
+		require.Failf(t, "", "SM Object needs to wait until sm.waitGetStateUntil")
```
bigbes, line 60: Please, add unit test that will check for this.

Moreover, that test may be flaky. If system scheduler will pause thread executing goroutine - we'll accept result sooner and it'll fail.

ledger-core/v2/virtual/object/object.go
bigbes: I'd suggest to move that part to separate function (or method)

PR #160, opened 2020-05-15
 PLAT-334: TestNewHostNetwork seems to be fixed on Windows 

afiskon:
TestNewHostNetwork seems to be fixed on Windows. CI checks were executed twice, all OK.


bigbes:
Nope, its not.
@afiskon lets revert it

PR #161, opened 2020-05-15
NOISSUE: update VCallRequest in CTMethod tests

ElSacramento:
**- What I did**

**- How I did it**

**- How to verify it**


ledger-core/v2/virtual/integration/method_test.go
```diff
@@ -323,11 +323,11 @@ func TestVirtual_Method_WithExecutor(t *testing.T) {
 			callResultPl := metaPayload.(*payload.Meta).Payload
 			callResultPlType, err := payload.UnmarshalType(callResultPl)
 			assert.NoError(t, err)
-			assert.Equal(t, payload.TypeVCallResult, callResultPlType)
+			assert.Equal(t, payload.TypeVStateRequest, callResultPlType)
```
bigbes, line 326: I've wrote a bad test, we need to prepare an object, before (so no VStateRequest is needed)
Later we should obtain VCallResult

ElSacramento: `Method_PrepareObject` doesn't prepare an object?

ElSacramento: Or I can use it?

ElSacramento: done

PR #169, opened 2020-05-17
INS-316: create SM object without reason

ruz:
there is no need in reason, SM object shares data, goes further into
sending VStateRequest, constructor or VStateReport just fill in data
immediately before request is sent


ruz:
incorrect task mark

PR #179, opened 2020-05-18
PLAT-280: SM for handle VDelegateRequestFinished

dgovorukhin:
**- What I did**
Added new SMVDelegatedRequestFinished for handle message. Also, added step to access shared data with object sm. update state and decrement pending counters. Release sync object for pending ordered operation if pending counters are 0.

**- How I did it**
 Write a code)
**- How to verify it**
virtual/integration/vdelegatedrequestfinished_test.go for 
Check that execute call will be called only after VDelegatedRequestFinished if we have pending ordered operation in VStateReport

https://insolar.atlassian.net/browse/PLAT-280


ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
```diff
@@ -0,0 +1,110 @@
...
+			objectRef,
+			state.Reference,
+			state.Prototype,
+			state.State,
+			state.Parent,
+		)
+		objectDescriptor = &desc
+	}
+
+	sharedObjectState := s.objectCatalog.Get(ctx, objectRef)
```
ivanshibitov, line 70: Get raises 'panic' if there is no such object.
What if DelegatedRequestFinished comes before any other requests/messages for this object?

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
```diff
@@ -0,0 +1,110 @@
...
+			ctx.Log().Trace(stateAlreadyExistsErrorMsg{
+				Reference: objectRef.String(),
+			})
+			return false
+		}
+
+		state.ActiveMutablePendingCount--
+
+		if state.ActiveMutablePendingCount < 0 {
+			//TODO how to handle?
```
ivanshibitov, line 83: Might be we should panic with IllegalState? No?

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
```diff
@@ -0,0 +1,110 @@
...
+	var objectDescriptor *descriptor.Object
+
+	if s.Payload.LatestState != nil {
+		state := s.Payload.LatestState
+		desc := descriptor.NewObject(
+			objectRef,
+			state.Reference,
+			state.Prototype,
+			state.State,
+			state.Parent,
```
ivanshibitov, line 65: This field is not set on sending side

ledger-core/v2/virtual/integration/vdelegatedrequestfinished_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	crBytes, err := cr.Marshal()
+	require.NoError(t, err)
+	vCallRequestMsg, err := wrapMsg(pulseNumber, crBytes)
+	require.NoError(t, err)
+
+	drfBytes, err := df.Marshal()
+	require.NoError(t, err)
+	vDelegateRequestFinishedMsg, err := wrapMsg(pulseNumber, drfBytes)
+	require.NoError(t, err)
```
ivanshibitov, line 107: may move this block to function?
something like
func makeMsg( t,  payload Marshaler ) *message.Message{
    raw, err := payload.Marshal()
  	require.NoError(t, err)
  	msg, err := wrapMsg(pulseNumber, raw)
  	require.NoError(t, err)
    return msg
}

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
dgovorukhin: @ivanshibitov As far I know we can not get VDelegatedRequestFinished before VStateReport.

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
dgovorukhin: @ivanshibitov While we do not have a pending table we can not properly handle this case. Talked with @cyraxred.

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
dgovorukhin: @ivanshibitov state.Parent? Or which one? What we will do?

ledger-core/v2/virtual/integration/vdelegatedrequestfinished_test.go
dgovorukhin: @ivanshibitov Fixed

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
ivanshibitov: yes. 
https://github.com/insolar/assured-ledger/blob/master/ledger-core/v2/virtual/execute/execute.go#L431
I didn't fill this field. Could you please do it?

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
```diff
@@ -0,0 +1,104 @@
...
+			})
+			return false
+		}
+
+		// Update object state.
+		if objectDescriptor != nil {
+			state.SetDescriptor(*objectDescriptor)
+		}
+
+		state.ActiveMutablePendingCount--
```
MikhailZhukov, line 85: Why we always decrease Mutable counter? Is it possible to send VDelegateRequestFinished after immutable execution?

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
ivanshibitov: Ideally, we have to be ready to such things. 
But only @cyraxred knows the truth

ledger-core/v2/virtual/integration/vdelegatedrequestfinished_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+	vCallRequestMsg, err := wrapMsg(pulseNumber, &cr)
+	require.NoError(t, err)
+
+	vDelegateRequestFinishedMsg, err := wrapMsg(pulseNumber, &df)
+	require.NoError(t, err)
+
+	server.SendMessage(ctx, vStateReportMsg)
+	aclSeq = append(aclSeq, "stateReportSend")
```
ivanshibitov, line 104: I would move all this strings ( "stateReportSend",  "callRequestSend" etc.) to contacts, at least because you use them twice

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
```diff
@@ -0,0 +1,104 @@
...
+
+		if state.ActiveMutablePendingCount == 0 {
+			// If we do not have pending ordered, release sync object.
+			ctx.CallBargeIn(state.AwaitPendingOrdered)
```
ivanshibitov, line 89: I would at least log it if CallBargeIn returns 'false'

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
dgovorukhin: @MikhailPopov  As far I know we gen Token only for ordered operation, and so we will receive VDelegatedRequestFinished the only for ordered operation from the previous executor. I double-check this question soon.

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
dgovorukhin: @ivanshibitov Can you help with text in the message? I do not know how to properly describe this situation.

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
dgovorukhin: @ivanshibitov Fixed

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
dgovorukhin: @MikhailZhukov Fixed

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
dgovorukhin: @ivanshibitov Fixed

ledger-core/v2/virtual/integration/vdelegatedrequestfinished_test.go
dgovorukhin: @ivanshibitov Fixed

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
ruz: @ivanshibitov is absolutely right, SM of this message even should wait for object to be ready like execute does. We can not rely on order of two messages. VPendingFinished can come faster than VStateReport.

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+		switch s.Payload.CallFlags.GetTolerance() {
+		case payload.CallIntolerable:
+			state.ActiveImmutablePendingCount--
+		case payload.CallTolerable:
+			state.ActiveMutablePendingCount--
+		}
+
+		if state.ActiveMutablePendingCount == 0 {
```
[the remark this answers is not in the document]
ruz: also, this condition can be true even if ActiveImmutablePendingCount was decremented. This means that after some point `if` block will be executed after each unordered pending is finished.

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
```diff
@@ -0,0 +1,111 @@
...
+
+		return true
+	}
+
+	switch sharedObjectState.PrepareAccess(setStateFunc).TryUse(ctx).GetDecision() {
+	case smachine.Passed:
+	case smachine.NotPassed:
+		return ctx.WaitShared(sharedObjectState.SharedDataLink).ThenRepeat()
+	default:
+		panic(throw.NotImplemented())
```
ruz, line 107: this should be `throw.Impossible`.

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
```diff
@@ -0,0 +1,111 @@
...
+			state.Parent,
+		)
+		objectDescriptor = &desc
+	}
+
+	sharedObjectState := s.objectCatalog.Get(ctx, objectRef)
+	setStateFunc := func(data interface{}) (wakeup bool) {
+		state := data.(*object.SharedState)
+		if !state.IsReady() {
+			ctx.Log().Trace(stateAlreadyExistsErrorMsg{
```
ruz, line 74: this is for sure is not correct log message. state is not ready and you log something completely different

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
dgovorukhin: @ruz Fixed

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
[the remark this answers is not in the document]
dgovorukhin: @ruz @sexyhamster Fixed

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
dgovorukhin: @ruz Fixed

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
dgovorukhin: @ruz Fixed

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
```diff
@@ -47,6 +49,32 @@ func (s *SMVDelegatedRequestFinished) GetStateMachineDeclaration() smachine.Stat
...
+		return ctx.WaitShared(s.objectSharedState.SharedDataLink).ThenRepeat()
+	case smachine.Impossible:
+		ctx.Log().Fatal("failed to get object state: already dead")
+	case smachine.Passed:
+	default:
+		panic(throw.NotImplemented())
+	}
+
+	if ctx.AcquireForThisStep(semaphoreReadyToWork).IsNotPassed() {
+		return ctx.Sleep().ThenRepeat()
```
sexyhamster, line 75: hmm, not sure who would wake up sm in this case. We have the same logic in execute @bigbes @cyraxred is this a proper behavior? I'm a bit confused

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
```diff
@@ -86,25 +113,27 @@ func (s *SMVDelegatedRequestFinished) stepProcess(ctx smachine.ExecutionContext)
 		case payload.CallIntolerable:
 			state.ActiveImmutablePendingCount--
```
sexyhamster: this one can underflow too

and, probably, it would be nice to log events that are not expeected, somethin like unexpected Pending finished for object %ref%. In case of error it would help

ledger-core/v2/virtual/handlers/vstateunavailable.go
```diff
@@ -61,6 +61,11 @@ type stateAlreadyExistsErrorMsg struct {
 	GotState  string
 }
 
+type stateIsNotReadyErrorMsg struct {
```
sexyhamster, line 64: this should be defined in file where it is used

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
dgovorukhin: @sexyhamster Fixed

ledger-core/v2/virtual/handlers/vstateunavailable.go
dgovorukhin: @sexyhamster Fixed

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
```diff
@@ -0,0 +1,164 @@
...
+				}
+			} else {
+				ctx.Log().Warn(unExpectedVDelegateRequestFinished{
+					Reference: objectRef.String(),
+					ordered:   true,
+				})
+			}
+		}
+
+		return true
```
ruz, line 105: I don't see why we should wakeup object's SM here

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
```diff
@@ -0,0 +1,164 @@
...
+	case smachine.NotPassed:
+		return ctx.WaitShared(s.objectSharedState.SharedDataLink).ThenRepeat()
+	case smachine.Impossible:
+		ctx.Log().Fatal("failed to get object state: already dead")
+	case smachine.Passed:
+	default:
+		panic(throw.NotImplemented())
+	}
+
+	if ctx.AcquireForThisStep(semaphoreReadyToWork).IsNotPassed() {
```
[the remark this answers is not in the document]
dgovorukhin: @ruz Fixed

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
dgovorukhin: @ruz Fixed

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
ruz: In short: SyncLinks have embedded mechanizm to wake up waiters. Was answered in a chat in details.

ledger-core/v2/virtual/handlers/handle_meta.go
```diff
@@ -104,6 +103,16 @@ func FactoryMeta(message *statemachine.DispatcherMessage) (pulse.Number, smachin
 			ctx.SetTracerID(traceID)
 			return &SMVStateUnavailable{Meta: payloadMeta, Payload: &pl}
 		}
+	case payload.TypeVDelegatedRequestFinished:
+		pl := payload.VDelegatedRequestFinished{}
```
cyraxred, line 107: @ruz @bigbes @sexyhamster 
btw there is no reason to use pointers for meta and messages:
1. Copy is cheap
2. SM can be created first, and then a message is unmarshalled into the field

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
```diff
@@ -0,0 +1,179 @@
...
+func (s *SMVDelegatedRequestFinished) awaitObjectReady(ctx smachine.ExecutionContext) smachine.StateUpdate {
+	if ctx.AcquireForThisStep(s.objectReadyToWork).IsNotPassed() {
+		return ctx.Sleep().ThenRepeat()
+	}
+
+	return ctx.Jump(s.stepProcess)
+}
+
+func (s *SMVDelegatedRequestFinished) stepProcess(ctx smachine.ExecutionContext) smachine.StateUpdate {
+	setStateFunc := func(data interface{}) (wakeup bool) {
```
cyraxred, line 97: I'd strongly recommend to use typed wrappers for shared data and avoid interface casting in SM code

ledger-core/v2/virtual/handlers/vdelegatedrequestfinished.go
```diff
@@ -0,0 +1,179 @@
...
+			})
+		}
+	case payload.CallTolerable:
+		if state.ActiveMutablePendingCount > 0 {
+			state.ActiveMutablePendingCount--
+
+			if state.ActiveMutablePendingCount == 0 {
+				// If we do not have pending ordered, release sync object.
+				if !ctx.CallBargeIn(state.AwaitPendingOrdered) {
+					ctx.Log().Trace("AwaitPendingOrdered BargeIn receive false")
```
cyraxred, line 150: This seems to be Warn

ledger-core/v2/virtual/integration/vdelegatedrequestfinished_test.go
```diff
@@ -0,0 +1,154 @@
...
+	select {
+	case res := <-requestIsDone:
+		aclSeq = append(aclSeq, res)
+		break
+	case <-time.After(20 * time.Second):
+		require.Failf(t, "", "timeout")
+	}
+
+	expSeq := []string{stateReportSend, callRequestSend, delegateRequestFinishSend, callResultReceive}
+	require.True(t, reflect.DeepEqual(expSeq, aclSeq))
```
cyraxred, line 131: Please use require.Equal - it compares slices properly and prints meaningful comparison

cyraxred: and what is a reason to compare expSeq and aclSeq when they are both statically generated within the same method?

ledger-core/v2/virtual/integration/vdelegatedrequestfinished_test.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func wrapMsg(pulseNumber pulse.Number, request payload.Marshaler) (*message.Message, error) {
+	bytes, err := request.Marshal()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to marshal request")
```
cyraxred, line 137: use throw.W, don't mix

ledger-core/v2/virtual/integration/vdelegatedrequestfinished_test.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/ThreeDotsLabs/watermill/message"
+	testwalletProxy "github.com/insolar/assured-ledger/ledger-core/v2/application/builtin/proxy/testwallet"
+	"github.com/insolar/assured-ledger/ledger-core/v2/insolar"
+	"github.com/insolar/assured-ledger/ledger-core/v2/insolar/gen"
+	"github.com/insolar/assured-ledger/ledger-core/v2/insolar/payload"
+	"github.com/insolar/assured-ledger/ledger-core/v2/instrumentation/inslogger"
+	"github.com/insolar/assured-ledger/ledger-core/v2/pulse"
+	"github.com/insolar/assured-ledger/ledger-core/v2/reference"
+	"github.com/insolar/assured-ledger/ledger-core/v2/virtual/integration/utils"
+	"github.com/pkg/errors"
```
cyraxred, line 18: github.com/pkg/errors 
shows that you are using wrong methods!

PR #183, opened 2020-05-19
PLAT-310 call requests list

AndreyBronin:
- What I did
add requests list in execute shared state for deduplication incoming requests algorithm
unit test in progress
add check for duplicate request

- How I did it
add `SawRequests map[reference.Global]struct{}` to shared state

- How to verify it
unit test


ledger-core/v2/virtual/object/object.go
```diff
@@ -46,6 +46,8 @@ type Info struct {
 
 	AwaitPendingOrdered smachine.BargeIn
 
+	SawRequests map[reference.Global]struct{}
```
sexyhamster, line 49: SeenRequests/KnownRequests. Not sure about the map, we have to not only be checking if requests is in list, but also we will need to find EarliestKnownPulse. Maybe OrderedMap would suit here better (it should already be in our codebase)

sexyhamster: but maybe just earliest/latest request field would do the job

AndreyBronin: @ruz do we need EarliestKnownPulse or ordered list for deduplication algorithm?

ruz: guys, if you look at the task and the story then you can find answer to these questions.

PR #188, opened 2020-05-19
NOISSUE: cleanup 

cyraxred:
**- What I did**
* moved types out of insolar package
* replaced insolar.PulseNumber with pulse.Number
* fix for trace.SetID to ignore same traceID


ledger-core/v2/insolar/pulsestor/memory.go
```diff
@@ -3,39 +3,39 @@
 // This material is licensed under the Insolar License version 1.0,
 // available at https://github.com/insolar/assured-ledger/blob/master/LICENSE.md.
 
-package pulse
+package pulsestor
```
sexyhamster, line 6: hmm i wonder why it is pulsestor, not pulsestore? :)

cyraxred: the name is ugly because I'd rather get rid of this package ;) but the change is already too big :(

PR #201, opened 2020-05-21
PLAT-295 pending table

aalekseev-ru:
**- What I did**
write PendingTable struct and add it to SM Object
**- How to verify it**
run tests


ledger-core/v2/virtual/object/object.go
```diff
@@ -55,6 +55,8 @@ type Info struct {
 	PotentialMutablePendingCount   uint8
 
 	objectState State
+
+	pendingTable pendingTable
```
sexyhamster, line 59: I wonder if we could avoid double mapping, we already have KnownRequests map, maybe we could have this maps in one place?

aalekseev-ru: missed that one, fixed

PR #205, opened 2020-05-22
PLAT-329: remove methodIsolation from heap

sexyhamster:
**- What I did**
removed methodIsolation from heap

**- How I did it**
- Reserved 0 value for flags, as discussed with @cyraxred 
- Added IsZero for flags and for contract.MethodIsolation struct
- Check methodIsolation fillin with IsZero

**- How to verify it**
run tests

### DIFF
--- a/ledger-core/v2/insolar/contract/virtual.go
+++ b/ledger-core/v2/insolar/contract/virtual.go
@@ -24,6 +24,10 @@ type MethodIsolation struct {
 	State        StateFlag
 }
 
+func (i MethodIsolation) IsZero() bool {
+	return i.Interference.IsZero() && i.State.IsZero()
+}
+
 // Method is a struct for Method and it's properties
 type Method struct {
 	Func      MethodFunc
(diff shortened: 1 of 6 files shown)


ledger-core/v2/insolar/contract/virtual.go
```diff
@@ -24,6 +24,10 @@ type MethodIsolation struct {
 	State        StateFlag
 }
 
+func (i MethodIsolation) IsZero() bool {
+	return i.Interference.IsZero() && i.State.IsZero()
```
cyraxred, line 28: Usually, check for one zeroable field would be enough

sexyhamster: somebody by mistake can call WithState, but forget about WithInterference.

So it's better to check both

cyraxred: dont mix
IsValid()
and
IsZero()

your implementation is for IsValid() ;)

PR #207, opened 2020-05-22
NOISSUE: fix test logs

cyraxred:
**- What I did**
* Proper handling of error and fatal logs in tests - test will fail
* Fix for error when replacing TraceID for SMObject

### DIFF
--- a/ledger-core/v2/virtual/integration/constructor_api_test.go
+++ b/ledger-core/v2/virtual/integration/constructor_api_test.go
@@ -18,9 +18,8 @@ import (
 
 func Test_API_Create(t *testing.T) {
 	t.Log("C4837")
-	ctx := inslogger.TestContext(t)
-
 	server := utils.NewServer(t)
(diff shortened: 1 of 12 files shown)


ledger-core/v2/virtual/integration/constructor_api_test.go
```diff
@@ -18,9 +18,8 @@ import (
 
 func Test_API_Create(t *testing.T) {
 	t.Log("C4837")
-	ctx := inslogger.TestContext(t)
-
 	server := utils.NewServer(t)
+	ctx := inslogger.TestContext(t)
```
sexyhamster, line 22: I was actually thinking of moving this before NewServer and adding ctx to NewServer

cyraxred: then
1. please don't access logger before  calling NewServer
2. make it same for all other tests (this was a reason for change - to unify across all tests)

PR #208, opened 2020-05-22
PLAT-295 penging table struct

aalekseev-ru:
**- What I did**
add pending table struct

**- How to verify it**
run tests

### DIFF
--- a/ledger-core/v2/virtual/object/pending_table.go
+++ b/ledger-core/v2/virtual/object/pending_table.go
@@ -1,0 +1,75 @@
+package object
+
+import (
+	"github.com/insolar/assured-ledger/ledger-core/v2/pulse"
+	"github.com/insolar/assured-ledger/ledger-core/v2/reference"
+)
+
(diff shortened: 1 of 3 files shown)


ledger-core/v2/virtual/object/pending_table.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+type pendingList struct {
+	oldestPulse pulse.Number
+	requests    map[reference.Global]*struct{
+		active bool
+	}
+}
+
+func newPendingTable() pendingTable {
```
afiskon, line 20: Private structures can't be used outside of package. Is it intentional?

afiskon: If not please consider adding a corresponding interface. Unit tests should test this interface.

aalekseev-ru: yeah, it was

PR #219, opened 2020-05-25
PLAT-351: activate pulse changing in network

MikhailZhukov:
**- What I did**
New Makefile target for run test_func many times with active pulse

**- How I did it**
* Allow to pulsard read run mode and pulse time from os ENV
* Add to Makefile.testing target for run with params `FUNCTEST_COUNT=100 PULSAR_ONESHOT=FALSE PULSAR_PULSETIME=3000`

**- How to verify it**
Run `make test_func_slow`

Add to pulsard ENV settings
* `PULSAR_ONESHOT` - if it have one of values `1`, `True`, `T`, `t`, `TRUE` pulsar will send one pulse to network and stop. Command line parameter `-o` set this mode too
* `PULSAR_PULSETIME` - allow override configuration file settings and set pulse time

### DIFF
--- a/ledger-core/v2/cmd/pulsard/main.go
+++ b/ledger-core/v2/cmd/pulsard/main.go
@@ -35,7 +35,7 @@ import (
 	"github.com/insolar/assured-ledger/ledger-core/v2/version"
 )
 
-const PulsarOneShotEnv = "PULSAR_ONE_SHOT"
+const PulsarOneShotEnv = "PULSAR.ONESHOT"
 
 type inputParams struct {
 	configPath string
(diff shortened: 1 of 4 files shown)


ledger-core/v2/cmd/pulsard/main.go
```diff
@@ -49,6 +52,10 @@ func parseInputParams() inputParams {
 	if err != nil {
 		fmt.Println("Wrong input params:", err.Error())
 	}
+	if oneShot, exists := os.LookupEnv(PulsarOneShotEnv); exists {
```
sexyhamster, line 55: please make it a part of configuration/or just get it from viper (viper not only parses file, it searches flags and ENV if proper ENV prefix is defined)

MikhailZhukov: done

PR #220, opened 2020-05-25
PLAT-372: Stop SMObject when it sent state and has no pending execs

MikhailZhukov:
**- What I did**
Send VStateReport once and stop SMObject when it sent state and has no pending executions

**- How I did it**
Increment counter when sent VStateReport and check it in stepWaitIndefinitely

**- How to verify it**
Run test

### DIFF
--- a/ledger-core/v2/virtual/object/object.go
+++ b/ledger-core/v2/virtual/object/object.go
@@ -121,6 +121,14 @@ func NewStateMachineObject(objectReference reference.Global) *SMObject {
 	}
 }
 
+type smObjectMigrateState int
+
+const (
+	stateWasNotSend smObjectMigrateState = iota
+	stateSent
+	readyToStop
+)
+
 type SMObject struct {
 	smachine.StateMachineDeclTemplate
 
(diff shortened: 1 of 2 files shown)


ledger-core/v2/virtual/object/object.go
```diff
@@ -255,6 +263,11 @@ func (sm *SMObject) createWaitPendingOrderedSM(ctx smachine.ExecutionContext) {
 }
 
 func (sm *SMObject) stepSendVStateReport(ctx smachine.ExecutionContext) smachine.StateUpdate {
+	if sm.migrationCount > 0 {
+		sm.migrationCount++
```
cyraxred, line 267: 1. misleading name
2. sm.migrationCount++ is called inside and outside of if - why not to call it before if?

MikhailZhukov: Fix name and counter increment

ledger-core/v2/virtual/object/object.go
```diff
@@ -238,6 +242,10 @@ func (sm *SMObject) stepReadyToWork(ctx smachine.ExecutionContext) smachine.Stat
 }
 
 func (sm *SMObject) stepWaitIndefinitely(ctx smachine.ExecutionContext) smachine.StateUpdate {
+	if !sm.hasPendingExecution() && sm.stateSentCount >= stateSentCounterToStop {
```
cyraxred, line 245: well, with new names then "stateSentCounterToStop" is meaningless

PR #229, opened 2020-05-26
PLAT-306: resend outgoing after migration

aalekseev-ru:
**- What I did**
add resend of outgoing after migration (pulse change)


ledger-core/v2/insolar/contract/virtual.go
```diff
@@ -79,3 +79,27 @@ const (
 func (f InterferenceFlag) IsZero() bool {
 	return f == 0
 }
+
+type SendResultFullFlag byte
```
sexyhamster, line 83: unused, the one that are used are in payload

aalekseev-ru: image
i named that one this way, it is unused now but will be used in future

ledger-core/v2/insolar/payload/vcallrequestflags.go
```diff
@@ -21,7 +21,7 @@ type SendResultFullFlag byte
 const (
 	SendResultDefault SendResultFullFlag = iota
 	SendResultFull
-	sendResultFullFlagCount = iota
+	sendResultFullFlagCount = iota - 1
```
afiskon, line 24: The name is wrong. What you meant was probably a number starting from zero.

ledger-core/v2/insolar/payload/vcallrequestflags_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+package payload
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
```
sexyhamster, line 14: you do need something like 
TestCallRequestFlags_BitStorageIsEngough
that ensures, that bit storage is enough to fit your flags

aalekseev-ru: https://github.com/insolar/assured-ledger/pull/229/files#diff-cf3d0de3817dffe38e53f19bf2bf656eR16-R23 why that is not enough?

sexyhamster: fair enough

ledger-core/v2/insolar/payload/vcallrequestflags.go
```diff
@@ -53,10 +51,7 @@ const (
 )
 
 func (f CallRequestFlags) WithSendResultFull(t SendResultFullFlag) CallRequestFlags {
-	if t == 0 {
-		panic(throw.IllegalValue())
-	}
-	if t > sendResultFullFlagCount {
+	if t > bitSendResultFullFlagCount {
```
afiskon, line 54: TWIMC - this looks awful. We compare SendResultFullFlag (type '**byte**') with the **number of bits** and depending on the result doing something. I wish we had better variable names and types, but apparently this is a subject for another PR.

aalekseev-ru: just used already approved naming https://github.com/insolar/assured-ledger/blob/master/ledger-core/v2/insolar/payload/vcallflags.go#L31

we can plan a new task about renaming and I can take it

PR #237, opened 2020-05-27
PLAT-366: fix NPE while sending VDelegatedFinished

ivanshibitov:
**- What I did**

**- How I did it**

**- How to verify it**


ledger-core/v2/virtual/execute/execute.go
```diff
@@ -581,6 +581,8 @@ func (s *SMExecute) setNewState(prototype reference.Global, memory []byte) func(
 			parentReference,
 		))
 
+		s.execution.ObjectDescriptor = state.Descriptor()
```
aalekseev-ru, line 584: can you add a test, that illustrate what you are fixing?

PR #242, opened 2020-05-28
PLAT-352: run functest many times on `master` with reduced pulse time

afiskon:
The proposed patch already catches some issues. I believe this is good enough for now.

### DIFF
--- a/ledger-core/v2/Makefile.testing
+++ b/ledger-core/v2/Makefile.testing
@@ -25,13 +25,10 @@ test_func: ## run functest FUNCTEST_COUNT times
 	PULSAR_PULSETIME=$(PULSAR_PULSETIME) \
 	CGO_ENABLED=1 $(GOTEST) -test.v $(TEST_ARGS) -tags "functest" ./application/functest -count=$(FUNCTEST_COUNT)
 
-.PHONY: test_func_very_slow
-test_func_very_slow: ## run functest 100 times with enabled pulse 3 sec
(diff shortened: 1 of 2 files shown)


ledger-core/v2/Makefile.testing
```diff
@@ -26,8 +26,9 @@ test_func: ## run functest FUNCTEST_COUNT times
 	CGO_ENABLED=1 $(GOTEST) -test.v $(TEST_ARGS) -tags "functest" ./application/functest -count=$(FUNCTEST_COUNT)
 
 .PHONY: test_func_slow
-test_func_slow: ## run functest 100 times with enabled pulse 3 sec
-	FUNCTEST_COUNT=100 PULSAR_ONESHOT=FALSE PULSAR_PULSETIME=3000 TEST_ARGS='-timeout 1200s -failfast' make test_func
+test_func_slow: # Run functests as long as practical to catch unexpected behavior during pulse change
+	pgrep insolard | xargs kill -9 # dirty hack to terminate all insolard processes
+	FUNCTEST_COUNT=100 PULSAR_ONESHOT=FALSE PULSAR_PULSETIME=5000 TEST_ARGS='-timeout 1200s -failfast' make test_func
```
sexyhamster, line 31: I guess that PULSAR_ONESHOT should be true. When func tests will run better on multiple pulses, then will enable them for both func and func_slow

sexyhamster: PULSAR_ONESHOT=$(PULSAR_ONESHOT) should do the work

afiskon: I'm very sorry, bit this strongly contradicts the Jira task.

PR #248, opened 2020-05-28
PLAT-309: request delegation token

AndreyBronin:
**- What I did**
Executor must request delegation token after migration if it have pending execution

**- How I did it**
Migration create subroutine which send VDelegatedCallRequest and wait response

**- How to verify it**
Run test

### DIFF
--- a/ledger-core/virtual/execute/delegatedtokenrequest.go
+++ b/ledger-core/virtual/execute/delegatedtokenrequest.go
@@ -21,7 +21,7 @@ import (
 	"github.com/insolar/assured-ledger/ledger-core/vanilla/throw"
 )
 
-type SMVDelegatedCallRequest struct {
+type SMDelegatedTokenRequest struct {
 	// input arguments
 	Meta           *payload.Meta
 	RequestPayload payload.VDelegatedCallRequest
@@ -35,39 +35,39 @@ type SMVDelegatedCallRequest struct {
 
 /* -------- Declaration ------------- */
 
-var dSMVDelegatedCallResultInstance smachine.StateMachineDeclaration = &dSMVDelegatedCallRequest{}
+var dSMDelegatedTokenRequestInstance smachine.StateMachineDeclaration = &dSMDelegatedTokenRequest{}
 
-type dSMVDelegatedCallRequest struct {
+type dSMDelegatedTokenRequest struct {
 	smachine.StateMachineDeclTemplate
 }
 
-func (*dSMVDelegatedCallRequest) InjectDependencies(sm smachine.StateMachine, _ smachine.SlotLink, injector *injector.DependencyInjector) {
-	s := sm.(*SMVDelegatedCallRequest)
+func (*dSMDelegatedTokenRequest) InjectDependencies(sm smachine.StateMachine, _ smachine.SlotLink, injector *injector.DependencyInjector) {
+	s := sm.(*SMDelegatedTokenRequest)
 
 	injector.MustInject(&s.pulseSlot)
 	injector.MustInject(&s.messageSender)
 }
 
-func (*dSMVDelegatedCallRequest) GetInitStateFor(sm smachine.StateMachine) smachine.InitFunc {
+func (*dSMDelegatedTokenRequest) GetInitStateFor(_ smachine.StateMachine) smachine.InitFunc {
 	return nil
 }
 
 /* -------- Instance ------------- */
 
-func (s *SMVDelegatedCallRequest) GetSubroutineInitState(_ smachine.SubroutineStartContext) smachine.InitFunc {
+func (s *SMDelegatedTokenRequest) GetSubroutineInitState(_ smachine.SubroutineStartContext) smachine.InitFunc {
 	return s.Init
 }
 
-func (s *SMVDelegatedCallRequest) GetStateMachineDeclaration() smachine.StateMachineDeclaration {
-	return dSMVDelegatedCallResultInstance
+func (s *SMDelegatedTokenRequest) GetStateMachineDeclaration() smachine.StateMachineDeclaration {
+	return dSMDelegatedTokenRequestInstance
 }
 
-func (s *SMVDelegatedCallRequest) Init(ctx smachine.InitializationContext) smachine.StateUpdate {
-	ctx.SetDefaultMigration(s.migrationDefault)
+func (s *SMDelegatedTokenRequest) Init(ctx smachine.InitializationContext) smachine.StateUpdate {
+	ctx.SetDefaultMigration(s.migration)
 	return ctx.Jump(s.stepRegisterBargeIn)
 }
 
-func (s *SMVDelegatedCallRequest) stepRegisterBargeIn(ctx smachine.ExecutionContext) smachine.StateUpdate {
+func (s *SMDelegatedTokenRequest) stepRegisterBargeIn(ctx smachine.ExecutionContext) smachine.StateUpdate {
 	bargeInCallback := ctx.NewBargeInWithParam(func(param interface{}) smachine.BargeInCallbackFunc {
 		res, ok := param.(*payload.VDelegatedCallResponse)
 		if !ok || res == nil {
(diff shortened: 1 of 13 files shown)


ledger-core/v2/virtual/handlers/vdelegatedcallrequest.go
```diff
@@ -0,0 +1,117 @@
+// Copyright 2020 Insolar Network Ltd.
```
sexyhamster, line 1: vdelegatedcallrequest? :O 
actually, we have different task for this handler PLAT-313

AndreyBronin: @sexyhamster  ths SM is for sending request, i put in in wrong package. Yes PLAT-311 is for handler

ledger-core/virtual/integration/delegated_finished_test.go
```diff
@@ -147,7 +147,7 @@ func TestVirtual_SendDelegatedFinished_IfPulseChanged_WithSideAffect(t *testing.
 
 		latestState := delegateFinishedMsg.LatestState
 		require.Equal(t, newRawWalletState, latestState.State)
-	case <-time.After(10 * time.Second):
+	case <-time.After(100 * time.Second):
```
ruz, line 150: return back 10 seconds

ledger-core/virtual/execute/delegatedtokenrequest_test.go
```diff
@@ -0,0 +1,129 @@
...
+			Caller: caller,
+			Callee: callee,
+		},
+		Meta: &payload.Meta{
+			Sender: caller,
+		},
+	}
+
+	var messageSender adapter.MessageSender = adapter.NewMessageSenderMock(mc).
+		PrepareAsyncMock.Set(
```
sexyhamster, line 70: in slotMachine we already have mocked sender, that could be prepared with
```
slotMachine.PrepareMockedMessageSender(mc)
```
and much easier to use 
```
slotMachine.MessageSender.SendTarget.Set(func(_ context.Context, msg payload.Marshaler, target reference.Global, _ ...messagesender.SendOption) error {
				res := msg.(*payload.VDelegatedCallResponse)
				// ensure that both times request is the same
				assert.Equal(t, caller, target)
				assert.Equal(t, expectedResponse, res)
				return nil
			})
```

MikhailZhukov: This way does not allow us to use ExecutionContext for GetPublishedGlobalAliasAndBargeIn and call it

ledger-core/virtual/execute/delegatedtokenrequest.go
```diff
@@ -0,0 +1,116 @@
...
+			panic(throw.IllegalValue())
+		}
+		s.response = res
+
+		return func(ctx smachine.BargeInContext) smachine.StateUpdate {
+			return ctx.WakeUp()
+		}
+	})
+
+	callOutgoing := gen.UniqueIDWithPulse(s.pulseSlot.PulseData().PulseNumber)
```
sexyhamster, line 83: @bigbes @ruz . I'm not sure, but this looks wrong, you already have outgoing call in VCallRequest.

ledger-core/virtual/execute/delegatedtokenrequest_test.go
sexyhamster: the main problem that you generate refIn, it should be static for object and request. So you won't even need to check bargeIn inside call, and just access it with slotMachine.GetPublishedGlobalAliasAndBargeIn, and ensure after step

ledger-core/virtual/execute/delegatedtokenrequest.go
MikhailZhukov: I think that this must be unique id for BargeIn and it can't be == outgoing call in VCallRequest

bigbes: Please, create alias type:
```
type DelegationTokenAwaitKey struct {
    Outgoing reference.Global
}
```

and use it for delegatonToken key. You can add pulse if you'll need separate processing for diefferent pulses

ledger-core/virtual/execute/delegatedtokenrequest_test.go
bigbes: I agree with @sexyhamster , that test construction looks very complicated. I'm sure it can be tested more straightforward way.

sexyhamster: So now, you don't need to get access to internal context.
You can check that bargeIn is registered after certain step of SMExcecute, get it from PulseSlot and call it in test body, so this code will be simplified

ledger-core/virtual/execute/delegatedtokenrequest.go
MikhailZhukov: done

ledger-core/virtual/execute/delegatedtokenrequest_test.go
MikhailZhukov: Done

ledger-core/virtual/execute/delegatedtokenrequest_test.go
```diff
@@ -0,0 +1,119 @@
...
+	slotMachine.Start()
+	defer slotMachine.Stop()
+	smWrapper := slotMachine.AddStateMachine(ctx, &smExecute)
+	{
+
+		slotMachine.RunTil(smWrapper.BeforeStep(smExecute.stepExecuteStart))
+		slotMachine.Migrate()
+		slotMachine.RunTil(smWrapper.AfterStep(smExecute.stepGetDelegationToken))
+		var smVDelegatedCallRequest SMDelegatedTokenRequest
+		slotMachine.RunTil(smWrapper.AfterStep(smVDelegatedCallRequest.stepSendRequest))
```
bigbes, line 101: SMDelegatedTokenRequest{}.stepSendRequest

MikhailZhukov: Done

ledger-core/virtual/execute/delegatedtokenrequest_test.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/insolar/assured-ledger/ledger-core/testutils/gen"
+	"github.com/insolar/assured-ledger/ledger-core/virtual/object"
+	"github.com/insolar/assured-ledger/ledger-core/virtual/testutils/slotdebugger"
+)
+
+func TestVirtual_CDelegatedCallRequest(t *testing.T) {
+	var (
+		mc       = minimock.NewController(t)
+		ctx      = inslogger.TestContext(t)
+		deadBeef = []byte{0xde, 0xad, 0xbe, 0xef}
```
bigbes, line 33: Isn't it better to use exported value from package? @sexyhamster

sexyhamster: hmm. Its better to create tokenService here and created expected token from arguments. Then you won't need deadBeef bytes

MikhailZhukov: That var private in test of other package. They do not have a general meaning and can be changed independently.

ledger-core/virtual/execute/delegatedtokenrequest_test.go
```diff
@@ -0,0 +1,118 @@
...
+		slotMachine.RunTil(smWrapper.BeforeStep(smExecute.stepExecuteStart))
+		slotMachine.Migrate()
+		slotMachine.RunTil(smWrapper.AfterStep(smExecute.stepGetDelegationToken))
+		slotMachine.RunTil(smWrapper.AfterStep(SMDelegatedTokenRequest{}.stepSendRequest))
+		slotLink, bargeInHolder := slotMachine.SlotMachine.GetPublishedGlobalAliasAndBargeIn(DelegationTokenAwaitKey{smExecute.execution.Outgoing})
+
+		require.False(t, slotLink.IsZero())
+		require.True(t, bargeInHolder.CallWithParam(&payload.VDelegatedCallResponse{
+			DelegationSpec:     payload.CallDelegationToken{Outgoing: smExecute.execution.Outgoing},
+			DelegatorSignature: deadBeef,
```
sexyhamster, line 106: DelegatorSignature is not used, ApproverSignature is inside DelegationSpec

ledger-core/virtual/execute/delegatedtokenrequest_test.go
sexyhamster: as far as I can see, there is no need in these bytes

MikhailZhukov: done

ledger-core/virtual/execute/delegatedtokenrequest_test.go
MikhailZhukov: remove it. done

PR #249, opened 2020-05-28
PLAT-311: earliest request pulse fillage

sexyhamster:
**- What I did**
- knownrequests should be tabled by tolerance
- earliestPulse for each path should be calculated according to tables

**- How to verify it**
run tests

### DIFF
--- a/ledger-core/v2/virtual/execute/execute.go
+++ b/ledger-core/v2/virtual/execute/execute.go
@@ -301,12 +301,13 @@ func (s *SMExecute) stepStartRequestProcessing(ctx smachine.ExecutionContext) sm
 	var objectDescriptor descriptor.Object
 
 	action := func(state *object.SharedState) {
-		if _, ok := state.KnownRequests[s.execution.Outgoing]; ok {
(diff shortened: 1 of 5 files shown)


ledger-core/v2/virtual/execute/execute.go
```diff
@@ -301,13 +301,14 @@ func (s *SMExecute) stepStartRequestProcessing(ctx smachine.ExecutionContext) sm
 	var objectDescriptor descriptor.Object
 
 	action := func(state *object.SharedState) {
-		if _, ok := state.KnownRequests[s.execution.Outgoing]; ok {
+		requestList := state.KnownRequests.GetList(s.execution.Isolation.Interference)
+		if requestList.Exist(s.execution.Outgoing) {
```
MikhailZhukov, line 305: Next line `requestList.Add(s.execution.Outgoing)` will return `false` if such record exist. You can combine add and check in one condition

sexyhamster: good point, will fix in next PR >_< missed your comment

PR #251, opened 2020-05-28
PLAT-362: Change proto generation for messages via protoc-gen-ins

dgovorukhin:
**- What I did**
Change proto generation for messages via protoc-gen-ins

**- How I did it**
- Change command in make file
- Change payload.proto and payload.r0.proto 
   - use insproto instead gogoproto
   - add option (insproto.id) instead Polymorph
   - add casttype for simple fields
   - add alias for message registration
- Remove PolymorphId

**- How to verify it**
run: make generate-protobuf (you should not see any changes in generated files)
run: make test_all

See more task details:
https://insolar.atlassian.net/browse/PLAT-362


ledger-core/v2/insolar/payload/payload.proto
```diff
@@ -3,14 +3,21 @@ syntax = "proto3";
 package payload;
 
 import "github.com/gogo/protobuf/gogoproto/gogo.proto";
+import "github.com/insolar/assured-ledger/ledger-core/v2/insproto/ins.proto";
+
+option (insproto.notation_all) = true;
+option (insproto.zeroable_all) = true;
```
sexyhamster, line 9: maybe I'm wrong, but we have couple of fields with are zeroable, all the rest should be required by default. Maybe we can turn this to false and explicitly say which fields are zeroable?

dgovorukhin: Make sense, I will check it. Thanks!

cyraxred: debatable ... if a field = 0, the why should we send bytes, denoting this field? receiver will do logical checks anyway

sexyhamster: hmmm, if it is kind of json "omitempty", then disregard my comment. In custom types it was  making *FieldType instead of FieldType (to be able to handle empty). I was thinking about that case.

cyraxred: well, if a field type has IsZero() method and zero value means missing value - like for references, then you don't need nullable as well. Just a matter of view point.

ledger-core/v2/virtual/integration/method_test.go
```diff
@@ -7,6 +7,7 @@ package integration
 
 import (
 	"context"
+	"github.com/insolar/assured-ledger/ledger-core/v2/rms"
```
bigbes, line 10: correct order of imports

dgovorukhin: =(

dgovorukhin: @bigbes Fixed

PR #252, opened 2020-05-29
NOISSUE: renamed all prototype occurences to class

ita-sammann:
Because we use not a prototype but class OOP model


ledger-core/v2/runner/requestresult/requestresult.go
```diff
@@ -58,8 +58,8 @@ func (s *RequestResult) SetAmend(object descriptor.Object, memory []byte) {
 	s.Memory = memory
 	s.ObjectStateID = object.StateID()
 
-	prototype, _ := object.Prototype()
-	s.ObjectImage = prototype
+	class, _ := object.Class()
+	s.ObjectImage = class
```
cyraxred, line 62: may be ClassImage then?

PR #257, opened 2020-06-01
PLAT-352: run separate checks on master

afiskon:
As Ruslan pointed out in comments to PLAT-352 it's difficult to find master-only runs in our GitHub repo. You should go to Actions -> linux-checks, use a filter by Branch.

Also apparently the originally proposed `if: ${{ github.event_name == 'push' }}` hack doesn't quite work as expected. Corresponding commands are not executed on master or pull requests at all.


afiskon:
@ita-sammann you realize we will stop running unit tests and functional tests on k8s in this case?

ita-sammann:
Then let's copy these checks to master-checks workflow. To have completely separate workflows for PR and master. I think it'll make our life easier in the future.

afiskon:
@ita-sammann OK

PR #280, opened 2020-06-03
PLAT-311: update knownRequests table upon finished requests processing

sexyhamster:
**- What I did**
update finished requests in known requests map so earliest request pulse would be appropriate

### DIFF
--- a/ledger-core/virtual/execute/execute.go
+++ b/ledger-core/virtual/execute/execute.go
@@ -534,7 +534,7 @@ func (s *SMExecute) stepSaveNewObject(ctx smachine.ExecutionContext) smachine.St
 	}
 
 	action = func(state *object.SharedState) {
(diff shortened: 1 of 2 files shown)


ledger-core/v2/virtual/execute/execute.go
```diff
@@ -608,9 +608,10 @@ func (s *SMExecute) stepSendDelegatedRequestFinished(ctx smachine.ExecutionConte
 	return ctx.Jump(s.stepSendCallResult)
 }
 
-func (s *SMExecute) decrementCounters(state *object.SharedState) {
+func (s *SMExecute) decrementCounters(state *object.SharedState, req reference.Global) {
```
MikhailZhukov, line 611: Maybe rename function because it do two things now

PR #283, opened 2020-06-05
PLAT-419: don't update objest.State if pulse was changed

ivanshibitov:
**- What I did**
Prevent updating of object shared state if pulse was changed

**- How I did it**

**- How to verify it**

### DIFF
--- a/ledger-core/virtual/execute/execute.go
+++ b/ledger-core/virtual/execute/execute.go
@@ -48,10 +48,11 @@ type SMExecute struct {
 	objectSharedState  object.SharedStateAccessor
 
 	// execution step
-	executionNewState *executionupdate.ContractExecutionStateUpdate
-	outgoingResult    []byte
-	deactivate        bool
-	run               *runner.RunState
+	executionNewState   *executionupdate.ContractExecutionStateUpdate
+	outgoingResult      []byte
+	deactivate          bool
+	run                 *runner.RunState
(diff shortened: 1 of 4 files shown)


ledger-core/virtual/execute/execute.go
```diff
@@ -543,16 +543,22 @@ func (s *SMExecute) stepSaveNewObject(ctx smachine.ExecutionContext) smachine.St
...
 		_, class, memory = executionNewState.Amend()
-		action = s.setNewState(class, memory)
+		action = s.getUpdateStateAction(class, memory)
 	case requestresult.SideEffectDeactivate:
 		panic(throw.NotImplemented())
 	default:
 		panic(throw.IllegalValue())
 	}
 
+	if s.migrationHappened {
```
sexyhamster, line 556: this should be done only if requestresult.SideEffectActivate or requestresult.SideEffectAmend
it is possible that nothing should be done with descriptor

sexyhamster: and test for this case should exist >_<

ivanshibitov: "it is possible that nothing should be done with descriptor" - What do you mean? I shouldn't change s.execution.ObjectDescriptor inside 'if' statement?

sexyhamster: My point is that we should not send latest state if state hasn't changed.

PR #296, opened 2020-06-08
NOISSUE: integration test server helpers and test fixes

sexyhamster:
subj

### DIFF
--- a/ledger-core/instrumentation/inslogger/test_helper.go
+++ b/ledger-core/instrumentation/inslogger/test_helper.go
@@ -12,7 +12,7 @@ import (
 )
 
 func NewTestLogger(target logcommon.TestingLogger, suppressTestError bool) log.Logger {
-	return NewTestLoggerExt(target, suppressTestError, "")
+	return NewTestLoggerExt(target, suppressTestError, "bilog")
 }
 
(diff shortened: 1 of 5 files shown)


ledger-core/instrumentation/inslogger/test_helper.go
```diff
@@ -24,6 +24,8 @@ func NewTestLoggerExt(target logcommon.TestingLogger, suppressTestError bool, ad
 	if adapter != "" {
 		logCfg.Adapter = adapter
 	}
+	logCfg.Level = logcommon.DebugLevel.String()
+	logCfg.Formatter = logcommon.TextFormat.String()
```
cyraxred, line 28: Why do you need text format?
JSON + build tag for text printer is much more readable

cyraxred: build tag for json to text (with filed sort etc) is 
convlogtxt

sexyhamster: when you ran/debug tests on localmachine, json is not very easy to parse, and with convlog it's a hell of a mess, when it's json mixed with text

PR #298, opened 2020-06-09
PLAT-477: text test run output and testrail cli update

sexyhamster:
**- What I did**
- use text output in all test runs
- test run outputs both to console and file
- removed -json test output
- added -v for verbose test logs

**- How to verify it**
see checks output

### DIFF
--- a/ledger-core/Makefile.testing
+++ b/ledger-core/Makefile.testing
@@ -15,7 +15,7 @@ test_unit: ## run all unit tests
 
 .PHONY: test_with_coverage
 test_with_coverage: ## run unit and slow tests with generation of coverage file
-	go-acc --covermode=count --output=coverage.tmp.out --ignore drafts ./... -- -tags=slowtest -count=1 $(TEST_ARGS) -json > "integration_test.log"
+	go-acc --covermode=count --output=coverage.tmp.out --ignore drafts ./... -- -v -tags=slowtest -count=1 $(TEST_ARGS) 2>&1 | tee "integration_test.log"
 	cat coverage.tmp.out | grep -v _mock.go > coverage.out
 	go tool cover -html=coverage.out -o coverage.html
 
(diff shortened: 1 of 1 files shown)


ledger-core/Makefile.testing
```diff
@@ -15,7 +15,7 @@ test_unit: ## run all unit tests
 
 .PHONY: test_with_coverage
 test_with_coverage: ## run unit and slow tests with generation of coverage file
-	go-acc --covermode=count --output=coverage.tmp.out --ignore ledger-v2 ./... -- -tags=slowtest -count=1 $(TEST_ARGS) -json > "integration_test.log"
+	go-acc --covermode=count --output=coverage.tmp.out --ignore ledger-v2 ./... -- -v -tags="slowtest convlogtxt" -count=1 $(TEST_ARGS) 2>&1 | tee "integration_test.log"
```
ElSacramento, line 18: testrail can't parse txt log

sexyhamster: then we can prepare output for testrail with [test2json](https://golang.org/cmd/test2json/)

I would really like to see actual test logs inside checks

ElSacramento: ok, add please into step `publish_integration_tests`

ruz: split make targets for CI from targets for developers

ita-sammann: It doesn't make sense to run `test_with_coverage` locally. So why split?

ruz: I didn't understand @sexyhamster 's reasons to do all this. Ignore my "split" comment.

sexyhamster: As discussed offline, I'll need to postprocess text test output to put minimal information to json for testrail. That gives us realtime output in CI run, and testrail information. I'll read more about testrail API and will figure out how to do that. If someone have free hands or idea how to make it quick, help would be much appreciated

PR #307, opened 2020-06-09
PLAT-311: fill token outgoing reference for matching on other side

sexyhamster:
subj

### DIFF
--- a/ledger-core/virtual/handlers/vdelegatedcallrequest.go
+++ b/ledger-core/virtual/handlers/vdelegatedcallrequest.go
@@ -194,7 +194,7 @@ func (s *SMVDelegatedCallRequest) stepProcessRequest(ctx smachine.ExecutionConte
 func (s *SMVDelegatedCallRequest) stepBuildResponse(ctx smachine.ExecutionContext) smachine.StateUpdate {
 	target := s.Meta.Sender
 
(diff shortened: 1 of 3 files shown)


ledger-core/virtual/handlers/vdelegatedcallrequest.go
```diff
@@ -194,7 +194,7 @@ func (s *SMVDelegatedCallRequest) stepProcessRequest(ctx smachine.ExecutionConte
 func (s *SMVDelegatedCallRequest) stepBuildResponse(ctx smachine.ExecutionContext) smachine.StateUpdate {
 	target := s.Meta.Sender
 
-	token := s.tokenService.GetCallDelegationToken(s.Meta.Sender, s.pulseSlot.PulseData().PulseNumber, s.Payload.Callee)
+	token := s.tokenService.GetCallDelegationToken(s.Payload.RequestReference, s.Meta.Sender, s.pulseSlot.PulseData().PulseNumber, s.Payload.Callee)
 
 	response := payload.VDelegatedCallResponse{
 		RefIn:          s.Payload.RefIn,
```
MikhailZhukov, line 200: RefIn must be s.Payload.RequestReference

sexyhamster: RefIn is RefIn, if we want to rename it, it should be done in one go, in both DelegatedCallRequest and DelegatedCallResponse

PR #309, opened 2020-06-09
PLAT-451: Access to conveyor cache

cyraxred:
**- What I did**
* Access to conveyor cache
* Change cache to store pulse.Range instead of pulse.Data
* Convenience functions to get prev pulse data

### DIFF
--- a/ledger-core/conveyor/conveyor.go
+++ b/ledger-core/conveyor/conveyor.go
@@ -103,6 +103,10 @@ type PulseConveyor struct {
 	stoppingChan <-chan struct{}
 }
 
+func (p *PulseConveyor) GetDataManager() *PulseDataManager {
+	return &p.pdm
+}
+
 func (p *PulseConveyor) AddDependency(v interface{}) {
 	p.slotMachine.AddDependency(v)
 }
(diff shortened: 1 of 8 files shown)


ledger-core/conveyor/conveyor.go
```diff
@@ -103,6 +103,10 @@ type PulseConveyor struct {
 	stoppingChan <-chan struct{}
 }
 
+func (p *PulseConveyor) GetDataManager() *PulseDataManager {
```
cyraxred, line 106: @ivanshibitov here is the first function you need

ledger-core/conveyor/pulse_manager.go
```diff
@@ -86,6 +87,21 @@ func (p *PulseDataManager) GetPresentPulse() (present pulse.Number, nearestFutur
 	return p._split(v)
 }
 
+func (p *PulseDataManager) GetPrevPulseRange() (pulse.Number, pulse.Range) {
```
cyraxred, line 90: @ivanshibitov here is the second function you need

ledger-core/conveyor/conveyor.go
ivanshibitov, line 106: It's not clear what is 'DataManager'. May be GetPulseDataManager?

cyraxred: well, there should be more data in it soon, hence the name

PR #337, opened 2020-06-15
NOISSUE: fix for misunderstanding

ElSacramento:
**- What I did**

**- How I did it**

**- How to verify it**


ElSacramento:
Decided with @bigbes that it's ok to use testwallet class for all tests.

PR #305, opened 2020-06-09
PLAT-354: Add delegationToken to outgoing messages in SMExecute

dgovorukhin:
**- What I did**
Added delegationToken to outgoing messages in SMExecute

**- How I did it**

Fill field in outgoing messages from field SMExecute

**- How to verify it**

**- Description for the changelog**

https://insolar.atlassian.net/browse/PLAT-354

PR #353, opened 2020-06-17
Plat 356 check message sender with token

ivanshibitov:
**- What I did**
Check Approver in token
Add tests for all messages

**- How I did it**

**- How to verify it**


ledger-core/insolar/payload/marker_auth.go
```diff
@@ -52,3 +52,9 @@ func (m *VDelegatedCallRequest) customSubject() (Reference, authSubjectMode) {
 func (m *VDelegatedCallResponse) customSubject() (Reference, authSubjectMode) {
 	return m.GetCallee(), hasAuthSubject
 }
+
+// It returns hasAuthSubject ( i.e. for current pulse ) since VDelegatedRequestFinished can come only with
```
bigbes, line 56: https://github.com/insolar/assured-ledger/pull/305/files#diff-dc883018a3c885bb7ee04be373b4ced3R56-R60

ledger-core/insolar/payload/proto_vnvn.r0.proto
```diff
@@ -202,7 +202,7 @@ message VDelegatedCallResponse {
 
     Reference Callee = 20;
     Reference CallIncoming = 21;
-    CallDelegationToken DelegationSpec = 22;
+    CallDelegationToken ResponseDelegationSpec = 22;
```
bigbes, line 205: why?

ivanshibitov: Because it has different meaning in comparison with other messages. Here, we send this token to node which ask it. In others messages we use it as proof that we are delegated to do some action. And also we need that VDelegatedCallResponse doesn't implement tokenHolder interface

PR #355, opened 2020-06-17
PLAT-476: BargeIn leak in SMDelegatedTokenRequest

bigbes:


ledger-core/conveyor/smachine/slot_aliases.go
```diff
@@ -55,13 +55,14 @@ func (a *slotAliases) RetainAll(parent *slotAliases, removeFn func(k interface{}
 			}
 		}
 	}
+
+	keys := a.keys[i:]
```
sexyhamster, line 59: ????

bigbes: !!!!

bigbes: There was a bug in a part where cleanup was processed, so it's @cyraxred fix for that problem

cyraxred: it is ;)

PR #358, opened 2020-06-18
NOISSUE: journal revamp

cyraxred:
**- What I did**
* Improved performance with Journal 
* Avoid possible deadlock or behavioral impact by Journal 
* Made replay (log history) optional and discard-able
* Stop journal goroutines on conveyor stop


ledger-core/virtual/integration/utils/server.go
```diff
@@ -174,11 +178,17 @@ func (s *Server) Init(ctx context.Context) {
...
 		panic(err)
 	}
-	s.Journal.Start()
 
 	s.pulseManager.AddDispatcher(s.virtual.FlowDispatcher)
 	s.IncrementPulseAndWaitIdle(ctx)
+}
+
+func (s *Server) StartRecording() {
+	s.StartRecordingExt(10_000, true)
```
bigbes, line 187: IMO discardOnOverflow should be false here, otherwise we'll silently timeout somewhere in the future

cyraxred: all Replay* method will panic on overflow of the recording buffer. 
Discard here only clears memory, but doesn't remove overflow status.

PR #362, opened 2020-06-18
refactoring: add server.InjectPayload helper function

ruz:


ledger-core/virtual/integration/utils/server.go
```diff
@@ -341,3 +341,8 @@ func (s *Server) ResetActiveConveyorFlag() {
 func (s *Server) WrapPayload(pl payload.Marshaler) *RequestWrapper {
 	return NewRequestWrapper(s.GetPulse().PulseNumber, pl).SetSender(s.caller)
 }
+
+func (s *Server) InjectPayload(ctx context.Context, pl payload.Marshaler) {
```
bigbes, line 430: Let's call it `SendPayload`, `Inject` looks like something low-level, but it's not

ruz: done

PR #341, opened 2020-06-16
PLAT-463: A.Foo -> B1.Bar, B2.Bar, B3.Bar

ElSacramento:
**- What I did**

**- How I did it**

**- How to verify it**

**- Description for the changelog**

PR #366, opened 2020-06-19
Plat 435 fix sequence inc

ElSacramento:
I want to merge it after https://github.com/insolar/assured-ledger/pull/341


ledger-core/virtual/integration/method_test.go
```diff
@@ -1232,3 +1230,237 @@ func TestVirtual_CallMultipleContractsFromContract_Ordered(t *testing.T) {
 
 	mc.Finish()
 }
+
+// call A.Foo -> B.Bar two times
+func TestVirtual_CallContractTwoTimes(t *testing.T) {
+	t.Log("C5183")
```
ruz, line 1236: Let's concentrate on testing Sequence field a bit. I think it's ok to drop passing requests to B object, e.g. avoid resending messages from A to B and generate response manually. This will cut test in half. Instead we can add more checks around sequence, for example do the following calls:

- call to A.Foo
-- A.Foo calls B.Some twice
- second call to A.Foo
-- A.Foo calls B.Some twice

check that Sequence on outgoing calls is 1, 2, 1, 2.

What do you think?

ElSacramento: done

ledger-core/virtual/integration/method_test.go
```diff
@@ -1191,3 +1192,201 @@ func TestVirtual_CallMultipleContractsFromContract_Ordered(t *testing.T) {
...
+	// add publish checker
+	{
+		server.PublisherMock.SetChecker(func(topic string, messages ...*message.Message) error {
+			require.Len(t, messages, 1)
+
+			pl, err := payload.UnmarshalFromMeta(messages[0].Payload)
+			require.NoError(t, err)
+
+			switch pl.(type) {
+			case *payload.VCallRequest:
```
sexyhamster, line 1315: server.PublisherMock.SetTypedChecker

ElSacramento: I don't want to resend msg to B, I need to make VCallResult by myself. 
If I'm right, it's impossible with TypedChecker.

ruz: it's possible and works just find

ElSacramento: done

ledger-core/virtual/integration/method_test.go
```diff
@@ -1529,3 +1527,190 @@ func TestVirtual_Method_Have_ObjectState(t *testing.T) {
...
+		Method_PrepareObject(ctx, server, payload.Ready, objectBGlobal)
+	}
+
+	// add ExecutionMocks to runnerMock
+	{
+		firstBuilder := execution.NewRPCBuilder(outgoingReasonFirst, objectAGlobal)
+		objectAExecutionFirstMock := runnerMock.AddExecutionMock(outgoingFirstCall.String())
+		objectAExecutionFirstMock.AddStart(
+			func(ctx execution.Context) {
+				t.Log("ExecutionStart [A.Foo] first call")
```
MikhailZhukov, line 1579: Do not use direct call t.Log. Use our logger as descried [here](https://insolar.atlassian.net/wiki/spaces/DEV/pages/1431568423/Core+v2+-+Logger+in+tests)

ElSacramento: removed

PR #369, opened 2020-06-19
WaitSignalsTimed test util: waits a few signals or timeouts

ruz:


ledger-core/virtual/testutils/utils.go
```diff
@@ -6,9 +6,44 @@
...
 }
+
+func WaitSignalsTimed(t *testing.T, timeout time.Duration, signals ...<-chan struct{}) {
+	cleanUp := make(chan struct{})
+	defer close(cleanUp)
+
+	var wg sync.WaitGroup
+	wg.Add(len(signals))
+	for _, c := range signals {
+		go func(signal <-chan struct{}) {
```
bigbes, line 29: nope, it's bad behaviour for `server.Journal.WaitAllAsyncCallsDone()`.
It MUST be checked only after `executeDone` is checked (so all async calls were send, but not all were processed)

bigbes: BTW when you call WailAllAsyncCallsDone - it's the point where async call count is >= 0, and if 0 is reached it returns true, so it'll fire
And it failed exact in that place, btw
image

ledger-core/virtual/integration/delegated_finished_test.go
```diff
@@ -239,19 +240,11 @@ func TestVirtual_SendDelegatedFinished_IfPulseChanged_Without_SideEffect(t *test
...
-			require.FailNow(t, "timeout")
-		}
-
-		select {
-		case <-server.Journal.WaitAllAsyncCallsDone():
-		case <-time.After(10 * time.Second):
-			require.FailNow(t, "timeout")
-		}
-	}
+	testutils.WaitSignalsTimed(
```
bigbes, line 243: Besides, I'd prefer common helper that returns `true`/`false`, use it in tests like
```
require.True(t, testutils.WaitSignalsTiimed(), "timeout")
```

ledger-core/virtual/testutils/utils.go
ruz: fixed `server.Journal.WaitAllAsyncCallsDone()` by separating calls, didn't realize that aspect at the moment of writing.

ledger-core/virtual/integration/delegated_finished_test.go
ruz: This is annoying boilerplate that my goal was to get rid off. I was even considering hiding duration in testutils package to avoid repeating 10*time.Seconds. May be name should be `testutils.RequireSignalsTimed`. What do you think?

fixed problem of locating point of error by using `t.Helper()` method. Now "timeout" error trace will show point in test, not in the utility function.

ledger-core/virtual/testutils/utils.go
```diff
@@ -6,9 +6,42 @@
...
+			}
+		}
+	}()
+
+	select {
+	case <-allGood:
+	case <-time.After(timeout):
+		require.FailNow(t, "timeout")
+	}
+}
```
bigbes, line 47:
```suggestion
func WaitSignalsTimed(t *testing.T, timeout time.Duration, signals ...<-chan struct{}) {
	cleanUp := make(chan struct{})
	defer close(cleanUp)

	timeoutChannel := time.After(timeout)

	for _, signal := range signals {
		select {
		case <-signal:
		case <-timeoutChannel:
			require.FailNow(t, "timeout")
		}
	}
}
```

Parallel waiting is completely unnecessary in that case, btw. And it simplifies code

ledger-core/virtual/integration/delegated_finished_test.go
bigbes: Sure.

ledger-core/virtual/testutils/utils.go
bigbes: 👍

ledger-core/virtual/testutils/utils.go
ruz: switched to sequential reading of channels and it does simplify code. However, in a few situations this may result in a deadlock. Deadlock may happen when signal channels are not buffered, signaled with message instead of `close()` and two channels are interconnected. First channel in the list waits second to proceed after signaling, but second is blocked on writing into channel. It's so rare case that I decided to go with simpler code.

PR #370, opened 2020-06-20
PLAT-493: request repeated twice in pulse, second when result is ready

ruz:
execution should happen once, VCallResult should be sent twice


ledger-core/virtual/integration/deduplication_test.go
```diff
@@ -253,3 +252,94 @@ func TestDeduplication_SecondCallOfMethodDuringExecution(t *testing.T) {
...
+	var firstResult *payload.VCallResult
+
+	typedChecker := server.PublisherMock.SetTypedChecker(ctx, mc, server)
+	typedChecker.VCallResult.Set(func(result *payload.VCallResult) bool {
+		if firstResult != nil {
+			require.Equal(t, firstResult, result)
+		} else {
+			firstResult = result
+		}
+
```
MariaVasilenko, line 322: we can add more checks?

PR #373, opened 2020-06-22
PLAT-548: Save execution results in SMObject

dgovorukhin:
**- What I did**
- Passed VCallResult from SMExecute to SMObject via shared state accessor

**- How I did it**
Wrote code =)

**- How to verify it**
See 
- TestSMExecute_VCallResultPassedToSMObject in execute_test.go

https://insolar.atlassian.net/browse/PLAT-548


ledger-core/virtual/object/object.go
```diff
@@ -56,6 +56,8 @@ type Info struct {
 	KnownRequests WorkingTable
 	// PendingTable holds requests that are known to be processed by other executors
 	PendingTable PendingTable
+	// ResultTable holds results that were executed
+	ResultTable ResultTable
```
ruz, line 60: I don't think we need ResultTable. I think this should be discussed with the team

dgovorukhin: @ruz Fixed

ledger-core/virtual/execute/execute.go
```diff
@@ -771,13 +774,12 @@ func (s *SMExecute) stepSendCallResult(ctx smachine.ExecutionContext) smachine.S
...
 
 func (s *SMExecute) stepFinishRequest(ctx smachine.ExecutionContext) smachine.StateUpdate {
-
 	if s.migrationHappened {
 		return ctx.Jump(s.stepSendDelegatedRequestFinished)
 	}
 
 	action := func(state *object.SharedState) {
-		state.FinishRequest(s.execution.Isolation, s.execution.Outgoing)
+		state.FinishRequest(s.execution.Isolation, s.execution.Outgoing, s.execution.Result)
```
sexyhamster, line 782: if you pass whole result, then you don't need s.execution.Isolation, you have CallFlags in result

PR #378, opened 2020-06-23
PLAT-394, PLAT-507: constructor call with bad class ref returns an error

ita-sammann:
Added error instead of panic when failed to find contract descriptor.
Added test checking that constructor calls with empty or random class ref return an error.

### DIFF
--- a/ledger-core/runner/runner.go
+++ b/ledger-core/runner/runner.go
@@ -25,6 +25,15 @@ import (
 	"github.com/insolar/assured-ledger/ledger-core/virtual/descriptor"
 )
 
+type ErrorDetail struct {
+	Type int
+}
+
+const (
+	DetailEmptyClassRef = iota
+	DetailBadClassRef
+)
+
 type UnmanagedService interface {
 	ExecutionStart(execution execution.Context) RunState
 	ExecutionContinue(run RunState, outgoingResult []byte)
@@ -173,7 +182,7 @@ func (r *DefaultService) executeMethod(
 
 	classReference, err := objectDescriptor.Class()
 	if err != nil {
-		return nil, errors.W(err, "couldn't get class reference")
+		return nil, errors.W(err, "couldn't get class reference", ErrorDetail{DetailEmptyClassRef})
 	}
 	if classReference.IsEmpty() {
 		panic(throw.IllegalState())
@@ -181,7 +190,7 @@ func (r *DefaultService) executeMethod(
 
 	classDescriptor, codeDescriptor, err := r.Cache.ByClassRef(ctx, classReference)
 	if err != nil {
-		return nil, errors.W(err, "couldn't get descriptors")
+		return nil, errors.W(err, "couldn't get descriptors", ErrorDetail{DetailBadClassRef})
 	}
 
 	codeExecutor, err := r.Manager.GetExecutor(codeDescriptor.MachineType())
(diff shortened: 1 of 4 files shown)


ledger-core/virtual/execute/execute.go
```diff
@@ -518,9 +517,19 @@ func (s *SMExecute) stepExecuteDecideNextStep(ctx smachine.ExecutionContext) sma
...
 		return ctx.Jump(s.stepSaveNewObject)
 	case execution.Error:
-		err := throw.W(newState.Error, "failed to execute request")
-		ctx.Log().Warn(err)
-		s.prepareExecutionError(err)
+		var detail runner.ErrorDetail
+		if throw.FindDetail(newState.Error, &detail) {
+			switch detail.Type {
+			case runner.DetailBadClassRef, runner.DetailEmptyClassRef:
+				s.prepareExecutionError(errors.New("bad class reference"))
```
sexyhamster, line 524: throw.E

ledger-core/runner/runner.go
```diff
@@ -173,15 +182,15 @@ func (r *DefaultService) executeMethod(
 
 	classReference, err := objectDescriptor.Class()
 	if err != nil {
-		return nil, errors.W(err, "couldn't get class reference")
+		return nil, errors.W(err, "couldn't get class reference", ErrorDetail{DetailEmptyClassRef})
```
sexyhamster, line 185: could you switch to direct throw usage?

ledger-core/virtual/execute/execute.go
```diff
@@ -518,9 +517,19 @@ func (s *SMExecute) stepExecuteDecideNextStep(ctx smachine.ExecutionContext) sma
 		// send VCallResult here
 		return ctx.Jump(s.stepSaveNewObject)
 	case execution.Error:
-		err := throw.W(newState.Error, "failed to execute request")
-		ctx.Log().Warn(err)
-		s.prepareExecutionError(err)
+		var detail runner.ErrorDetail
+		if throw.FindDetail(newState.Error, &detail) {
```
cyraxred, line 521: just in case - it can also be a one-liner

if d := runner.ErrorDetail{}; throw.FindDetail(newState.Error, &d) {
...
}

ledger-core/runner/runner.go
```diff
@@ -173,15 +182,15 @@ func (r *DefaultService) executeMethod(
...
 	}
 	if classReference.IsEmpty() {
 		panic(throw.IllegalState())
 	}
 
 	classDescriptor, codeDescriptor, err := r.Cache.ByClassRef(ctx, classReference)
 	if err != nil {
-		return nil, errors.W(err, "couldn't get descriptors")
+		return nil, errors.W(err, "couldn't get descriptors", ErrorDetail{DetailBadClassRef})
```
sexyhamster, line 193: I was thinking about 
```
var  ErrBadClassRef = ErrorDetail{DetailBadClassRef}
throw.W(err, "couldn't get descriptors", ErrBadClassRef)
```
so it would be easier to use.

cyraxred: imho, this is better not to do - to avoid any sharing of details among different error instances

ledger-core/runner/runner.go
cyraxred: yes, errors -> throw
this "errors" name is a remnant of previous rewrite to remove /x/errors

ledger-core/virtual/execute/execute.go
ita-sammann: Thanks, fixed. But had to use `new(runner.ErrorDetail)`

ledger-core/virtual/execute/execute.go
ita-sammann: Fixed, thanks!

ledger-core/runner/runner.go
ita-sammann: Replaced `errors` package with `throw` everywhere in `runner.go`

PR #382, opened 2020-06-23
PLAT-533 + PLAT-485: bugfix outgoing token resend

sexyhamster:
**- What I did**
- proper token for outgoing resending
- test that verifies proper token


ledger-core/virtual/integration/method_test.go
```diff
@@ -1704,3 +1707,192 @@ func TestVirtual_CallContractTwoTimes(t *testing.T) {
...
+			assert.Equal(t, payload.CTMethod, request.CallType)
+			assert.Equal(t, uint32(1), request.CallSequence)
+			assert.Equal(t, outgoingCallRef, request.CallReason)
+			assert.Equal(t, payload.BuildCallFlags(barIsolation.Interference, barIsolation.State), request.CallFlags)
+
+			assert.Equal(t, firstPulse, request.CallOutgoing.Pulse())
+
+			if callCount == 0 {
+				assert.Equal(t, payload.CallDelegationToken{}, request.DelegationSpec)
+				server.IncrementPulseAndWaitIdle(ctx)
```
ElSacramento, line 1839: increment must be before VCallResult response 
`пульс меняется до того как пришел VCallResult от контракта B `

sexyhamster: well, pulse changes after VCallRequest sent and before VCallResult received.

ledger-core/virtual/integration/method_test.go
```diff
@@ -1704,3 +1707,192 @@ func TestVirtual_CallContractTwoTimes(t *testing.T) {
...
+			},
+			&execution.Update{
+				Type:   execution.Done,
+				Result: requestresult.New([]byte("finish B.Foo"), objectBGlobal),
+			},
+		)
+
+		runnerMock.AddExecutionMock("Bar").AddStart(
+			func(ctx execution.Context) {
+				t.Log("ExecutionStart [A.Bar]")
```
ElSacramento, line 1802: remove t.log or change it  
https://insolar.atlassian.net/wiki/spaces/DEV/pages/1431568423/Core+v2+-+Logger+in+tests

sexyhamster: switched to initialized logger from test context

ledger-core/virtual/integration/method_test.go
```diff
@@ -661,7 +672,7 @@ func TestVirtual_Call_UnorderedMethod_From_UnorderedMethod(t *testing.T) {
 
 	runnerMock.AddExecutionMock("Bar").AddStart(
 		func(ctx execution.Context) {
-			t.Log("ExecutionStart [A.Bar]")
+			logger.Debug("ExecutionStart [A.Bar]")
 		},
```
ruz, line 676: come on, do this kind of cleanups via NOISSUE branches

ledger-core/virtual/integration/method_test.go
```diff
@@ -911,7 +926,7 @@ func TestVirtual_CallMethodFromConstructor_Unordered(t *testing.T) {
 
 		runnerMock.AddExecutionMock("Foo").AddStart(
```
ruz, line 927: I think that it is better to drop mock of executing B. Just don't resend VCallRequest, check token in the message and generate manually VCallResponse.

ledger-core/virtual/integration/method_test.go
```diff
@@ -1704,3 +1726,194 @@ func TestVirtual_CallContractTwoTimes(t *testing.T) {
...
+	Method_PrepareObject(ctx, server, payload.Ready, objectBGlobal)
+
+	// add ExecutionMocks to runnerMock
+	{
+		outgoingCall := execution.NewRPCBuilder(outgoingCallRef, objectBGlobal).
+			CallMethod(objectAGlobal, classA, "Bar", []byte("123")).
+			SetInterference(barIsolation.Interference).
+			SetIsolation(barIsolation.State)
+		objectBExecutionMock := runnerMock.AddExecutionMock("Foo")
+		objectBExecutionMock.AddStart(
```
ruz, line 1798: I think that it is better to drop mock of executing B. Just don't resend VCallRequest, check token in the message and generate manually VCallResponse.

sexyhamster: not sure about this approach, It'll require more work, if something around VCallResponse will change.

ledger-core/virtual/integration/method_test.go
```diff
@@ -1704,3 +1726,194 @@ func TestVirtual_CallContractTwoTimes(t *testing.T) {
...
+				assert.Equal(t, payload.CallDelegationToken{}, request.DelegationSpec)
+				server.IncrementPulseAndWaitIdle(ctx)
+				callCount++
+				// request will be sent in previous pulse
+				// ommit sending
+				return false
+			}
+
+			service := authentication.NewService(ctx, server.JetCoordinatorMock)
+			origToken := service.GetCallDelegationToken(
```
ruz, line 1868: No-no, you don't use platform code for checks. You should check every field using test constants.

sexyhamster: could you explain why? Result of GetCallDelegationToken output should be tested elsewhere. If GetCallDelegationToken logic will change, why do you want this test to fail?

ledger-core/virtual/integration/method_test.go
ruz: What if interface of execute adapter would change?

How this B's part is relevant to delegation token sent by call from A to B? As far as I know in this test the token is not checked by recipient. So it's irrelevant part. As I said many times messages are our boundaries for integration tests.

ledger-core/virtual/integration/method_test.go
```diff
@@ -1704,3 +1726,194 @@ func TestVirtual_CallContractTwoTimes(t *testing.T) {
...
+		typedChecker.VDelegatedCallResponse.SetResend(true)
+		typedChecker.VDelegatedRequestFinished.SetResend(true)
+		typedChecker.VCallRequest.Set(func(request *payload.VCallRequest) bool {
+			assert.Equal(t, objectAGlobal, request.Callee)
+			assert.Equal(t, objectBGlobal, request.Caller)
+			assert.Equal(t, []byte("123"), request.Arguments)
+			assert.Equal(t, payload.CTMethod, request.CallType)
+			assert.Equal(t, uint32(1), request.CallSequence)
+			assert.Equal(t, outgoingCallRef, request.CallReason)
+			assert.Equal(t, payload.BuildCallFlags(barIsolation.Interference, barIsolation.State), request.CallFlags)
```
ruz, line 1854: we test code around obtaining delegation token and attaching it to external calls. it's very strange that so many fields of VCallRequest are checked, but **ZERO** fields of VDelegatedCallRequest, VDelegatedCallResponse are checked. Also, VDelegatedRequestFinished is not checked for presence of the token.

sexyhamster: This test tests token that is attached to outgoing call from contract. VDelegatedCallRequest and VDelegatedCallResponse should be tested elsewhere with their own tests. From my point of view, testing all the messages in all the tests is a waste of time.

ledger-core/virtual/integration/method_test.go
sexyhamster: switched to message instead of execution. Probably we need to rethink most of the tests in this file. Most of them actually execute more than one request.

sexyhamster:
I strongly disagree, you want this test to emulate a lot of the system logic and test not only this case but also bunch of other stuff. Other suff (VStateReport/VDelegatedCallRequest/VDelegatedCallResponse) should have their own test cases. The thing with naturally - add another test case and create a ticket for that.

sexyhamster:
OK, as discussed - I removed all the parts of the system that is outside of test scope. 
Double pulse is discribed in different case - C5205, not sure that this could be done via table test
Added one more case for execution that lasts 2 pulses.

ledger-core/virtual/integration/outgoing_test.go
```diff
@@ -0,0 +1,221 @@
...
+
+	Method_PrepareObject(ctx, server, payload.Ready, objectAGlobal)
+
+	// add ExecutionMocks to runnerMock
+	{
+		outgoingCall := execution.NewRPCBuilder(outgoingCallRef, objectAGlobal).
+			CallMethod(objectBGlobal, classB, "Bar", []byte("123")).
+			SetInterference(barIsolation.Interference).
+			SetIsolation(barIsolation.State)
+		objectBExecutionMock := runnerMock.AddExecutionMock("Foo")
```
ElSacramento, line 94: objectAExecutionMock

ledger-core/virtual/integration/outgoing_test.go
```diff
@@ -0,0 +1,221 @@
...
+			assert.Equal(t, expectedToken, finished.DelegationSpec)
+			return false
+		})
+		typedChecker.VCallRequest.Set(func(request *payload.VCallRequest) bool {
+			assert.Equal(t, objectBGlobal, request.Callee)
+			assert.Equal(t, objectAGlobal, request.Caller)
+
+			assert.Equal(t, firstPulse, request.CallOutgoing.Pulse())
+			if callCount == 0 {
+				assert.Equal(t, payload.CallDelegationToken{}, request.DelegationSpec)
```
ElSacramento, line 163: you can use `request.DelegationSpec.IsZero()` (just proposal)

ledger-core/virtual/integration/outgoing_test.go
```diff
@@ -0,0 +1,221 @@
...
+				CallType:        request.CallType,
+				CallFlags:       request.CallFlags,
+				CallAsOf:        request.CallAsOf,
+				Caller:          request.Caller,
+				Callee:          request.Callee,
+				CallOutgoing:    request.CallOutgoing,
+				ReturnArguments: []byte("finish B.Bar"),
+			})
+
+			return false // resend
```
ElSacramento, line 183: wrong comment

ledger-core/virtual/integration/outgoing_test.go
```diff
@@ -0,0 +1,221 @@
...
+	{
+		typedChecker.VStateReport.Set(func(report *payload.VStateReport) bool {
+			assert.Equal(t, firstPulse, report.OrderedPendingEarliestPulse)
+			assert.Equal(t, int32(1), report.OrderedPendingCount)
+			return false
+		})
+		typedChecker.VDelegatedCallRequest.Set(func(request *payload.VDelegatedCallRequest) bool {
+			assert.Zero(t, request.DelegationSpec)
+			assert.Equal(t, objectAGlobal, request.Callee)
+			assert.Equal(t, payload.BuildCallFlags(fooIsolation.Interference, fooIsolation.State), request.CallFlags)
```
ElSacramento, line 142: why do you check flags? I do think this test isn't about isolation.

ledger-core/virtual/integration/outgoing_test.go
sexyhamster: when it fails - here it will tell you that expectations is empty CallDelegationToken, but received CallDelegationToken with values. assert.True will assert that wanted true, got false, which is less verbose

sexyhamster: well, on the second thought, I could switch to assert.Zero

ledger-core/virtual/integration/outgoing_test.go
sexyhamster: removed extra checks.

ledger-core/virtual/integration/outgoing_test.go
sexyhamster: fixed

ledger-core/virtual/integration/outgoing_test.go
```diff
@@ -0,0 +1,195 @@
...
+			if callCount == 0 {
+				assert.Zero(t, request.DelegationSpec)
+				server.IncrementPulseAndWaitIdle(ctx)
+				callCount++
+				// request will be sent in previous pulse
+				// omit sending
+				return false
+			}
+
+			assert.Equal(t, expectedToken, request.DelegationSpec)
```
ElSacramento, line 151: check repeat flag, please
`Repeated Call`

ledger-core/virtual/integration/outgoing_test.go
```diff
@@ -0,0 +1,195 @@
...
+	server.SendMessage(ctx, msg)
+
+	// wait for all calls and SMs
+	testutils.WaitSignalsTimed(t, 10*time.Second, executeDone)
+	testutils.WaitSignalsTimed(t, 10*time.Second, server.Journal.WaitAllAsyncCallsDone())
+
+	assert.Equal(t, 2, typedChecker.VCallRequest.Count())
+	assert.Equal(t, 1, typedChecker.VCallResult.Count())
+	assert.Equal(t, 1, typedChecker.VStateReport.Count())
+	assert.Equal(t, 1, typedChecker.VDelegatedCallRequest.Count())
```
ElSacramento, line 192: forget about VDelegatedRequestFinished

ledger-core/virtual/integration/outgoing_test.go
sexyhamster: done

PR #383, opened 2020-06-23
PLAT-388: Implemented SMCallSummary 

dgovorukhin:
**- What I did**
Create a new SMCallSummary with will store all VCallResult for a specific pulse and will be stopped after migration to next pulse.

**- How I did it**
- New package callsummary
- Move PendingTable and WorkingTable to package table (need for resolve cycle imports)
- Add a new step in SMObject and SMExecute for publishing data to SMCallSummary after migration

**- How to verify it**
New test virtual/object/object_publish_callsummary_test.go

https://insolar.atlassian.net/browse/PLAT-388


ledger-core/virtual/object/object.go
```diff
@@ -398,14 +404,74 @@ func (sm *SMObject) createWaitPendingOrderedSM(ctx smachine.ExecutionContext) {
...
+func (sm *SMObject) stepMigration(ctx smachine.MigrationContext) smachine.StateUpdate {
 	if sm.GetState() == Unknown {
 		ctx.Log().Trace("SMObject migration happened when object is not ready yet")
 		return ctx.Stop()
 	}
 
 	ctx.UnpublishAll()
 
+	// store migration pulse
+	sm.MigratePulseNumber = sm.pulseSlot.PulseData().PulseNumber
```
sexyhamster, line 416: you don't need to store it in object. You always can call sm.pulseSlot.PulseData().PulseNumber. It will not change

ledger-core/virtual/execute/execute.go
```diff
@@ -824,6 +824,19 @@ func (s *SMExecute) stepSendCallResult(ctx smachine.ExecutionContext) smachine.S
...
+		if ok {
+			callSummaryAccessor.Prepare(func(sharedCallSummary *object.SharedCallSummary) {
+				ref := s.execution.Outgoing
+				requests, ok := sharedCallSummary.Requests.GetObjectsKnownRequests(ref)
+				if ok {
+					requests.GetList(s.execution.Isolation.Interference).Finish(ref, s.execution.Result)
+				}
+			}).TryUse(ctx)
+		}
+
```
ruz, line 839: above block is not correct, it doesn't account situation when SM Object was not able to pass in data. what I mean is that request can reach this step after migration faster than object's finalizer.

also, this code doesn't check result of TryUse. Doesn't wait for shared and so on...

ledger-core/virtual/object/callsummary.go
```diff
@@ -0,0 +1,97 @@
...
+func (v CallSummarySharedStateAccessor) PrepareAndWakeUp(fn func(*SharedCallSummary)) smachine.SharedDataAccessor {
+	return v.PrepareAccess(func(data interface{}) bool {
+		fn(data.(*SharedCallSummary))
+		return true
+	})
+}
+
+func GetSummarySMSharedAccessor(
+	ctx smachine.ExecutionContext,
+	summaryKey SummarySharedKey,
```
sexyhamster, line 91: I think it would be easier to use this package, if knowabout SummarySharedKey would be inside this package. As a user you can only pass PulseNumber

ledger-core/virtual/object/object.go
```diff
@@ -398,14 +404,74 @@ func (sm *SMObject) createWaitPendingOrderedSM(ctx smachine.ExecutionContext) {
...
+
+	sdl := ctx.Share(&sm.SummaryDone, 0)
+
+	if !ctx.Publish(SummarySyncKey{
```
sexyhamster, line 420: there should be wrapper for easy access to this synckey

ledger-core/virtual/object/object.go
```diff
@@ -398,14 +404,74 @@ func (sm *SMObject) createWaitPendingOrderedSM(ctx smachine.ExecutionContext) {
...
+		objectRef:   sm.Reference,
+		pulseNumber: sm.MigratePulseNumber,
```
sexyhamster, line 422: pulseNumber could be omited. This syncKey lives in one pulse slot. There should not be situation when we'll have two keys for one object in one pulseSlot

ledger-core/virtual/object/object.go
dgovorukhin: Fixed

ledger-core/virtual/execute/execute.go
dgovorukhin: Fixed

ledger-core/virtual/object/object.go
dgovorukhin: Fixed

ledger-core/virtual/callsummary/callsummary.go
```diff
@@ -0,0 +1,125 @@
...
+
+import (
+	"github.com/insolar/assured-ledger/ledger-core/conveyor/smachine"
+	"github.com/insolar/assured-ledger/ledger-core/pulse"
+	"github.com/insolar/assured-ledger/ledger-core/reference"
+	"github.com/insolar/assured-ledger/ledger-core/vanilla/injector"
+	"github.com/insolar/assured-ledger/ledger-core/virtual/tables"
+)
+
+type pulseNumber = pulse.Number
```
sexyhamster, line 18: why do you need an alias?

ledger-core/virtual/execute/execute.go
```diff
@@ -737,6 +739,56 @@ func (s *SMExecute) stepSaveNewObject(ctx smachine.ExecutionContext) smachine.St
 	return ctx.Jump(s.stepSendCallResult)
 }
 
+func (s *SMExecute) stepPublishCallSummaryData(ctx smachine.ExecutionContext) smachine.StateUpdate {
+	syncAccessor, ok := callsummary.GetSummarySMSyncAccessor(ctx, s.execution.Object)
+
+	if ok {
+		var syncLink smachine.SyncLink
```
sexyhamster, line 746: I'd move this to separate step

ledger-core/virtual/execute/execute.go
```diff
@@ -737,6 +739,56 @@ func (s *SMExecute) stepSaveNewObject(ctx smachine.ExecutionContext) smachine.St
...
+
+		if ctx.AcquireForThisStep(syncLink).IsNotPassed() {
+			return ctx.Sleep().ThenRepeat()
+		}
+	}
+
+	callSummaryAccessor, ok := callsummary.GetSummarySMSharedAccessor(ctx, s.pulseSlot.PulseData().PulseNumber)
+
+	if ok {
+		action := func(sharedCallSummary *callsummary.SharedCallSummary) {
```
sexyhamster, line 769: and this could be separate step

ledger-core/virtual/callsummary/callsummary.go
dgovorukhin: Fixed

ledger-core/virtual/execute/execute.go
dgovorukhin: Fixed

ledger-core/virtual/execute/execute.go
dgovorukhin: Fixed

ledger-core/virtual/callsummary/callsummary.go
```diff
@@ -0,0 +1,123 @@
...
+
+func BuildSummarySyncKey(objectRef reference.Global) SummarySyncKey {
+	return SummarySyncKey{objectRef: objectRef}
+}
+
+type SummarySharedKey struct {
+	pulseNumber pulse.Number
+}
+
+func BuildSummarySharedKey(pulse pulse.Number) SummarySharedKey {
```
bigbes, line 30: I don't like names `BuildSummarySharedKey` and `BuildSummarySyncKey `; I can't get difference of what they're doing outside of context. Can you, at least, add commentary? Or finding better name to describe their nature? 
Like `SummarySyncKey` is for temporary use, until `SummarySharedKey` is stored. Or something like that.

dgovorukhin: Sure, I may add a comment. Why we have func BuildReportKey(object reference.Global) ReportKey?

bigbes: No, problem not in function for building key, but that i cannot understand difference between two that types of keys

ledger-core/virtual/tables/objects_request_table.go
```diff
@@ -0,0 +1,30 @@
+// Copyright 2020 Insolar Network Ltd.
+// All rights reserved.
+// This material is licensed under the Insolar License version 1.0,
+// available at https://github.com/insolar/assured-ledger/blob/master/LICENSE.md.
+
+package tables
```
bigbes, line 6: please, try to avoid naming of package in plural form

bigbes: also i'd be glad if it can give more information - what table? table is pretty common name for data structure

ledger-core/virtual/execute/execute.go
bigbes: It's not pushed yet?

ledger-core/virtual/object/object.go
```diff
@@ -428,8 +435,63 @@ func (sm *SMObject) stepMigrate(ctx smachine.MigrationContext) smachine.StateUpd
...
+	sdlCallSummarySync := ctx.Share(&sm.SummaryDone, 0)
+
+	if !ctx.Publish(callsummary.BuildSummarySyncKey(sm.Reference), sdlCallSummarySync) {
+		ctx.Log().Warn(struct {
+			*log.Msg  `txt:"failed to publish call summary sync key"`
+			Reference string
+		}{
+			Reference: sm.Reference.String(),
+		})
+		return ctx.Error(fmt.Errorf("failed to publish call summary sync key"))
```
bigbes, line 447: What is the reason to use fmt.Errorf here? avoid it; use throw.New instead

ledger-core/virtual/object/object.go
```diff
@@ -428,8 +435,63 @@ func (sm *SMObject) stepMigrate(ctx smachine.MigrationContext) smachine.StateUpd
 		return ctx.Error(fmt.Errorf("failed to publish state report"))
 	}
 
+	sdlCallSummarySync := ctx.Share(&sm.SummaryDone, 0)
```
bigbes, line 438: Can that new code be extracted from here into callsummary package?

ledger-core/virtual/object/object.go
```diff
@@ -428,8 +435,63 @@ func (sm *SMObject) stepMigrate(ctx smachine.MigrationContext) smachine.StateUpd
...
+func (sm *SMObject) stepPublishCallSummary(ctx smachine.ExecutionContext) smachine.StateUpdate {
+	pulseNumber := sm.pulseSlot.PulseData().PulseNumber
+	summarySharedStateAccessor, ok := callsummary.GetSummarySMSharedAccessor(ctx, pulseNumber)
+
+	if !ok {
+		ctx.InitChild(func(ctx smachine.ConstructionContext) smachine.StateMachine {
+			return callsummary.NewStateMachineCallSummary(pulseNumber)
+		})
+
+		summarySharedStateAccessor, _ = callsummary.GetSummarySMSharedAccessor(ctx, pulseNumber)
```
bigbes, line 462: check for `!ok` here

ledger-core/virtual/object/object.go
```diff
@@ -428,8 +435,63 @@ func (sm *SMObject) stepMigrate(ctx smachine.MigrationContext) smachine.StateUpd
...
+		shared.Requests.AddObjectRequests(sm.Reference, sm.KnownRequests)
+
+		if !ctx.Unpublish(callsummary.BuildSummarySyncKey(sm.Reference)) {
+			ctx.Log().Warn(struct {
+				*log.Msg  `txt:"failed to unpublish call summary sync key"`
+				Reference string
+			}{
+				Reference: sm.Reference.String(),
+			})
+			ctx.Error(fmt.Errorf("failed to unpublish call summary sync key"))
```
bigbes, line 475: Same for fmt.Errorf here

ledger-core/virtual/object/object_migration_test.go
```diff
@@ -66,12 +67,17 @@ func TestSMObject_MigrationCreateStateReport_IfStateMissing(t *testing.T) {
 		LogMock.Return(smachine.Logger{}).
 		ShareMock.Return(smachine.NewUnboundSharedData(&report)).
 		PublishMock.Set(func(key interface{}, data interface{}) (b1 bool) {
```
bigbes, line 69: That publishMockFn became bigger and more complicated.
Can you extract it before migrationCtx is created and, also, cheeck that ReportKey AND SummarySyncKey both called. Also default - we should check other cases.

ledger-core/virtual/object/object_migration_test.go
```diff
@@ -109,21 +115,29 @@ func TestSMObject_MigrationCreateStateReport_IfStateIsEmptyAndNoCounters(t *test
 		UnpublishAllMock.Return().
 		ShareMock.Set(
 		func(data interface{}, flags smachine.ShareDataFlags) (s1 smachine.SharedDataLink) {
-			require.IsType(t, &payload.VStateReport{}, data)
+			switch d := data.(type) {
+			case payload.VStateReport:
+				require.IsType(t, &payload.VStateReport{}, d)
```
bigbes, line 120: oof, bad

ledger-core/virtual/object/object_migration_test.go
```diff
@@ -109,21 +115,29 @@ func TestSMObject_MigrationCreateStateReport_IfStateIsEmptyAndNoCounters(t *test
...
 			return sharedData
 		}).
 		PublishMock.Set(
 		func(key interface{}, data interface{}) (b1 bool) {
-			refKey := finalizedstate.BuildReportKey(smObject.Reference)
-			require.Equal(t, refKey, key)
-			require.Equal(t, sharedData, data)
+			switch d := data.(type) {
+			case finalizedstate.ReportKey:
+				refKey := finalizedstate.BuildReportKey(smObject.Reference)
```
bigbes, line 130: make that checks consistent with each other

ledger-core/virtual/object/object_migration_test.go
```diff
@@ -109,21 +115,29 @@ func TestSMObject_MigrationCreateStateReport_IfStateIsEmptyAndNoCounters(t *test
...
+				require.Equal(t, refKey, key)
+				require.Equal(t, sharedData, data)
+			case callsummary.SummarySyncKey:
+				assert.Equal(t, callsummary.BuildSummarySyncKey(smObject.Reference), d)
```
bigbes, line 134: can you check shared data here?

ledger-core/virtual/object/object_migration_test.go
```diff
@@ -148,12 +162,18 @@ func TestSMObject_MigrationCreateStateReport_IfStateEmptyAndCountersSet(t *testi
 		LogMock.Return(smachine.Logger{}).
 		ShareMock.Return(smachine.NewUnboundSharedData(&report)).
 		PublishMock.Set(func(key interface{}, data interface{}) (b1 bool) {
-		assert.Equal(t, finalizedstate.BuildReportKey(report.Object), key)
 		assert.NotNil(t, data)
```
bigbes, line 165: extract that function too, before migrationCtx

ledger-core/virtual/object/object_publish_callsummary_test.go
```diff
@@ -0,0 +1,149 @@
...
+		smObject.migrate(migrationCtx)
+	}
+
+	{
+		scs := callsummary.SharedCallSummary{
+			Requests: tables.NewObjectRequestTable(),
+		}
+		sdl := smachine.NewUnboundSharedData(&scs)
+
+		firsCall := true
```
bigbes, line 107: firstCall

ledger-core/virtual/object/object_publish_callsummary_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+		executionCtx := smachine.NewExecutionContextMock(mc).
+			GetPublishedLinkMock.Set(func(key interface{}) (s1 smachine.SharedDataLink) {
```
bigbes, line 110: please, rewrite that function with something like:
```
var rv smachine.SharedDataLink
switch key {
case callsummary.BuildSummarySharedKey(pulseNumber):
    if !firstCall {
       firstCall = false
    } else {
        rv = sdl
    }
default:
    t.Fatal(...)
}
```

bigbes: and extract it before executionCtx creation

ledger-core/virtual/execute/execute.go
dgovorukhin: Pushed

dgovorukhin: stepAwaitSMCallSummary
stepPublishDataCallSummary

ledger-core/virtual/object/object_migration_test.go
dgovorukhin: Explain, please

ledger-core/virtual/object/object.go
dgovorukhin: Sure, we can, for what?

ledger-core/virtual/object/object.go
dgovorukhin: Fixed

ledger-core/virtual/callsummary/callsummary.go
dgovorukhin: Added comments for keys

ledger-core/virtual/object/object.go
dgovorukhin:
```
ctx.InitChild(func(ctx smachine.ConstructionContext) smachine.StateMachine {
		p.initChildCtx(ctx, objectReference)

		return NewStateMachineObject(objectReference)
	})

	accessor, _ := p.TryGet(ctx, objectReference)
```

Example from catalog.go, should we change here too?

ledger-core/virtual/object/object_migration_test.go
sexyhamster: hmm. Why do you need to check for type after type assertion? Do you not want to test type assertion?

ledger-core/virtual/object/object.go
dgovorukhin: Fixed

ledger-core/virtual/object/object_migration_test.go
dgovorukhin: Fixed

ledger-core/virtual/object/object_migration_test.go
dgovorukhin: Fixed

ledger-core/virtual/object/object_migration_test.go
dgovorukhin: Fixed

ledger-core/virtual/object/object_publish_callsummary_test.go
dgovorukhin: Fixed

ledger-core/virtual/object/object_publish_callsummary_test.go
dgovorukhin: Fixed

ledger-core/virtual/object/object_migration_test.go
sexyhamster: and if d is supposed to be &payload.VStateReport{}, then type assertion should be 
```
case *payload.VStateReport:
```

dgovorukhin: We need to skip other types check

sexyhamster: I mean after type assert - you are pretty sure what type d is. What require supposed to do?

ledger-core/virtual/tables/objects_request_table.go
dgovorukhin: Why we use
```
type WorkingTable struct {
	lists map[contract.InterferenceFlag]*WorkingList
}
```
 what does table mean in this case? Same comment for this structure.

I may change the package name to 'store'. It will be more clear?

ledger-core/virtual/object/object_migration_test.go
dgovorukhin: Fixed

ledger-core/virtual/object/object_migration_test.go
dgovorukhin: Fixed

ledger-core/virtual/object/object.go
bigbes: si

ledger-core/virtual/tables/objects_request_table.go
bigbes: You can ask @cyraxred in that case, I don't know good/clear name for that package. 

`tables` is not a good name, anyway.

ledger-core/virtual/object/object.go
dgovorukhin: why I can not ignore ok here? Same as object catalog case

ledger-core/virtual/tables/objects_request_table.go
dgovorukhin: Fixed, rename the package to 'callregistry'

ledger-core/virtual/object/object.go
dgovorukhin: Fixed + fixed in catalog.go

ledger-core/virtual/execute/execute.go
```diff
@@ -720,6 +722,69 @@ func (s *SMExecute) stepSaveNewObject(ctx smachine.ExecutionContext) smachine.St
...
+}
+
+func (s *SMExecute) stepPublishDataCallSummary(ctx smachine.ExecutionContext) smachine.StateUpdate {
+	callSummaryAccessor, ok := callsummary.GetSummarySMSharedAccessor(ctx, s.pulseSlot.PulseData().PulseNumber)
+
+	if ok {
+		action := func(sharedCallSummary *callsummary.SharedCallSummary) {
+			ref := s.execution.Outgoing
+			requests, ok := sharedCallSummary.Requests.GetObjectsKnownRequests(s.execution.Object)
+			if ok {
```
sexyhamster, line 759: if !ok panic?

dgovorukhin: mm, maybe your right. If we finish operation we should have object result in this map.

sexyhamster: Maybe put this all into method on SharedCallSummary, so internal structure would not leak outside?

ledger-core/virtual/callregistry/objects_request_table.go
```diff
@@ -0,0 +1,34 @@
...
+
+package callregistry
+
+import (
+	"github.com/insolar/assured-ledger/ledger-core/reference"
+)
+
+// store request per object by object reference.
+type ObjectsRequestsTable struct {
+	knownRequests map[reference.Global]*CallResults
```
sexyhamster, line 14: If you will add method like 
```
func (ort *ObjectsRequestsTable) PutResults(ref, results)
```

then you don't need *CallResults over here. I think it would be better

ledger-core/virtual/callsummary/callsummary.go
```diff
@@ -0,0 +1,125 @@
...
+
+import (
+	"github.com/insolar/assured-ledger/ledger-core/conveyor/smachine"
+	"github.com/insolar/assured-ledger/ledger-core/pulse"
+	"github.com/insolar/assured-ledger/ledger-core/reference"
+	"github.com/insolar/assured-ledger/ledger-core/vanilla/injector"
+	"github.com/insolar/assured-ledger/ledger-core/virtual/callregistry"
+)
+
+// key for publish synclink, need for await publish call results after migration for specific object
```
sexyhamster, line 18: godoc style is prefered 
// SummarySyncKey is used ...

ledger-core/virtual/callsummary/callsummary.go
```diff
@@ -0,0 +1,125 @@
...
+type SummarySyncKey struct {
+	objectRef reference.Global
+}
+
+func BuildSummarySyncKey(objectRef reference.Global) SummarySyncKey {
```
sexyhamster, line 23: Comment over here would be much appreciated. Because users will work with this function and doc is now relateed to SummarySyncKey

ledger-core/virtual/execute/execute.go
dgovorukhin: Fixed

ledger-core/virtual/callregistry/objects_request_table.go
sexyhamster: * and also allocates less

dgovorukhin: CallResult need for wrap map in knownRequests.
```
knownRequests map[reference.Global]*CallResults
```
vs
```
knownRequests map[reference.Global]map[reference.Global]CallSummary
```

ledger-core/virtual/callsummary/callsummary.go
dgovorukhin: Fixed

ledger-core/virtual/callsummary/callsummary.go
dgovorukhin: Fixed

ledger-core/virtual/callregistry/objects_request_table.go
dgovorukhin: @sexyhamster Please check new version

ledger-core/virtual/callregistry/objects_request_table.go
```diff
@@ -0,0 +1,34 @@
...
+
+package callregistry
+
+import (
+	"github.com/insolar/assured-ledger/ledger-core/reference"
+)
+
+// ObjectsResultCallRegistry is used for store results for all processed requests per object by object reference.
+type ObjectsResultCallRegistry struct {
+	objects map[reference.Global]*ObjectCallResults
```
sexyhamster, line 14: what I meant is it could be map[reference.Global]ObjectCallResults

ledger-core/virtual/execute/execute.go
```diff
@@ -720,6 +722,75 @@ func (s *SMExecute) stepSaveNewObject(ctx smachine.ExecutionContext) smachine.St
...
+	return ctx.Jump(s.stepPublishDataCallSummary)
+}
+
+func (s *SMExecute) stepPublishDataCallSummary(ctx smachine.ExecutionContext) smachine.StateUpdate {
+	callSummaryAccessor, ok := callsummary.GetSummarySMSharedAccessor(ctx, s.pulseSlot.PulseData().PulseNumber)
+
+	if ok {
+		action := func(sharedCallSummary *callsummary.SharedCallSummary) {
+			ref := s.execution.Outgoing
+			requests, ok := sharedCallSummary.Requests.GetObjectsCallResults(s.execution.Object)
```
sexyhamster: I'd prefer this logic to be hidden inside sharedCallSummary object. 
something like 
```
func (orcr *ObjectsResultCallRegistry) SaveResult(ref, result) bool
```

ledger-core/virtual/callregistry/objects_request_table.go
```diff
@@ -0,0 +1,63 @@
...
+	callSummary, ok := requests.CallResults[reqRef]
+	if !ok {
+		// if we not have summary in resultMap, it mean we do not go through stepStartRequestProcessing,
+		// and we can't have a result here
+		panic(throw.Impossible())
+	}
+
+	if callSummary.Result != nil {
+		// we should not have result because migration was before we publish result in object
+		panic(throw.Impossible())
```
sexyhamster, line 59: throw.InvalidState() and all above (it is possible, but something went wrong)

ledger-core/virtual/execute/execute.go
dgovorukhin: Fixed

ledger-core/virtual/callsummary/callsummary.go
sexyhamster: for godoc 
// BuildSummarySyncKey builds SummarySyncKey, which is used...

ledger-core/virtual/callregistry/objects_request_table.go
dgovorukhin: done

ledger-core/virtual/callsummary/callsummary.go
dgovorukhin: done

PR #385, opened 2020-06-23
PLAT-494: test deduplication of method calls using VFindCallRequest

ruz:


ledger-core/virtual/integration/deduplication_test.go
```diff
@@ -318,3 +321,525 @@ func TestDeduplication_SecondCallOfMethodAfterExecution(t *testing.T) {
...
+		{
+			name: "no pendings, findCall message, found call, has result",
+
+			expectFindRequestMessage: true,
+			findRequestStatus:        payload.FoundCall,
+			findRequestHasResult:     true,
+
+			expectResultMessage: true,
+		},
+
```
aalekseev-ru, line 387: there must be test "no pendings, findCall message, found call, no result",

ruz: I don't think so. I believe that it's situation that shouldn't happen in release 0. If previous executor knows about request then executor either produced result or request is in pending executions, but we don't have pendings, so it has result. In future releases such case may happen once we start "forgetting" results.

ruz: I would rather see panic Impossible if we got "found without result", pendings table is complete and request is not in the table, but I didn't write test for that.

ledger-core/virtual/integration/deduplication_test.go
```diff
@@ -318,3 +321,525 @@ func TestDeduplication_SecondCallOfMethodAfterExecution(t *testing.T) {
...
+func (s *deduplicateMethodUsingPrevVETest) getClass() reference.Global {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	return s.class
+}
+
+func (s *deduplicateMethodUsingPrevVETest) getOutgoingRef() reference.Global {
+	callee := s.getObject() // TODO: FIXME: should be caller
```
aalekseev-ru, line 661: is there jira issue for this?

ruz: Nope.

ledger-core/virtual/integration/deduplication_test.go
aalekseev-ru: I can make panic in my code and add a test for that case

ledger-core/virtual/integration/deduplication_test.go
```diff
@@ -318,3 +321,525 @@ func TestDeduplication_SecondCallOfMethodAfterExecution(t *testing.T) {
...
+func (s *deduplicateMethodUsingPrevVETest) waitMessagePublications(
+	ctx context.Context,
+	t *testing.T,
+	expected int,
+) {
+	t.Helper()
+	if !s.server.PublisherMock.WaitCount(expected, 10*time.Second) {
+		require.Fail(t, "timeout waiting for messages on publisher")
+	}
+	require.Equal(t, expected, s.server.PublisherMock.GetCount())
```
ElSacramento, line 831: You can use it
```
	require.Equal(t, 4, typedChecker.VCallRequest.Count())
	require.Equal(t, 2, typedChecker.VCallResult.Count())
```

ruz: done

ledger-core/virtual/integration/deduplication_test.go
ruz: created

ledger-core/virtual/integration/deduplication_test.go
ElSacramento: function `waitMessagePublications` can be removed now, am I right?

ruz: yep, thanks, done

PR #386, opened 2020-06-24
NOISSUE: (R1) Any holder + Excerpt

cyraxred:


ledger-core/rms/proto_gen.go
```diff
@@ -7,4 +7,4 @@ package rms
 
 // `go:generate` command below is here just for information. It should not be used by go generate.
 // Actual generation is in the Makefile.
-// go:generate protoc -I=. -I=$GOPATH/src --ins_out=./ proto_*.proto
+//go:generate protoc -I=. -I=$GOPATH/src --ins_out=./ proto_*.proto
```
sexyhamster, line 10: either comment above is outdated, or this string has to have space before

cyraxred: that's artifact of my local work - pls ignore ;) WIP

ledger-core/rms/any.go
```diff
@@ -0,0 +1,104 @@
...
+	case thatValue == nil:
+		return p.value == nil
+	case p.value == nil:
+		return false
+	}
+
+	if eq, ok := thatValue.(interface{ Equal(that interface{}) bool}); ok {
+		return eq.Equal(p.value)
+	}
+	return false
```
ruz, line 103: should we panic here? Uncomparable values?

ledger-core/rms/any_lazy.go
```diff
@@ -0,0 +1,214 @@
...
+
+func (p *AnyLazy) Get() LazyValue {
+	if vv, ok := p.value.(LazyValue); ok {
+		return vv
+	}
+	return LazyValue{}
+}
+
+func (p *AnyLazy) Set(v GoGoSerializable) {
+	p.value = v
```
ruz, line 30: is it impossible to check for LazyValue when we set? Set/Get sequence returns silently empty LazyValue for incorrect type. I would have placed some runtime asserts.

ledger-core/rms/any.go
cyraxred: it is a general notation that Equal should return false then

ledger-core/rms/any_lazy.go
cyraxred: well a lazy container is usually asymmetrical - set is for not-serialized value, and read for serialized / lazy one. Will correct naming to make the difference explicit

cyraxred: done
https://github.com/insolar/assured-ledger/pull/386/commits/c9461a556377796c6d9ea2d51d618c6aa19302fb

PR #389, opened 2020-06-24
PLAT-397: bad contract method call returns error instead of panic

AndreyBronin:
**- What I did**
- update C4976 TestVirtual_BadMethod_WithExecutor
- provide VCallResult with a wrapped error instead of panic


ledger-core/virtual/integration/method_test.go
```diff
@@ -92,14 +110,19 @@ func TestVirtual_BadMethod_WithExecutor(t *testing.T) {
 			Callee:              objectGlobal,
 			CallSiteDeclaration: class,
 			CallSiteMethod:      "random",
-			CallOutgoing:        server.RandomLocalWithPulse(),
+			CallOutgoing:        objectLocal,
```
ruz, line 113: generate another Local reference for outgoing request. don't use object's

PR #393, opened 2020-06-25
PLAT-555: refactor tests and add test for call unordered B from unordered A

MariaVasilenko:
**- What I did**
- fix syntax t.Skip for valid integation testrail with jira
- refactor tests to table
- add test for call unordered B from unordered A(WIP)

**- How I did it**

**- How to verify it**


MariaVasilenko:
github not view latest commits

PR #399, opened 2020-06-25
PLAT-398: Add integration test to call non-existing constructor

IgorMarnat:
**- What I did**
-Added integration test C4977 for calling non-existing constructor
-Improved integration test C5180 (calling existing constructor) by addition of checking that it returns no error (otherwise it can be called with any name of constructor and still return success)


ledger-core/virtual/integration/constructor_test.go
```diff
@@ -548,3 +553,53 @@ func TestVirtual_CallConstructorFromConstructor(t *testing.T) {
...
+		Callee:         testwallet.GetClass(),
+		CallSiteMethod: "NotExistingConstructorName",
+		CallOutgoing:   outgoing,
+		Arguments:      insolar.MustSerialize([]interface{}{}),
+	}
+
+	typedChecker := server.PublisherMock.SetTypedChecker(ctx, mc, server)
+
+	typedChecker.VCallResult.Set(func(res *payload.VCallResult) bool {
+		require.Equal(t, res.Callee, objectRef)
```
MariaVasilenko, line 587: please, fix require syntax:
`func Equal(t TestingT, expected interface{}, actual interface{}, msgAndArgs ...interface{}) {`
move expected and actual values

IgorMarnat: Good point, done.

PR #400, opened 2020-06-26
PLAT-395: bad object call returns error instead of panic

AndreyBronin:
**- What I did**

update C4974 TestVirtual_Method_WithExecutor_ObjectIsNotExist
provide VCallResult with an error instead of panic


ledger-core/virtual/execute/execute.go
```diff
@@ -235,13 +235,14 @@ func (s *SMExecute) stepWaitObjectReady(ctx smachine.ExecutionContext) smachine.
 			panic(throw.NotImplemented())
 		}
 	} else if objectState != object.HasState {
-		panic(throw.E("no state on object after readyToWork", struct {
+		s.prepareExecutionError(throw.E("no state on object after readyToWork", struct {
```
sexyhamster, line 238: maybe better to say error - "object is not ready"

ledger-core/virtual/integration/method_test.go
```diff
@@ -174,16 +175,49 @@ func TestVirtual_Method_WithExecutor(t *testing.T) {
...
 	server.IncrementPulse(ctx)
 
 	var (
 		objectLocal  = server.RandomLocalWithPulse()
 		objectGlobal = reference.NewSelf(objectLocal)
+		outgoing     = server.RandomLocalWithPulse()
 	)
 
-	Method_PrepareObject(ctx, server, payload.Ready, objectGlobal)
+	Method_PrepareObject(ctx, server, payload.Ready, reference.NewSelf(server.RandomLocalWithPulse()))
```
sexyhamster, line 190: there is no need in this prepare object (it creates random object, which is not relevant to the test)

ledger-core/virtual/integration/method_test.go
```diff
@@ -174,16 +175,49 @@ func TestVirtual_Method_WithExecutor(t *testing.T) {
...
+		State           object.State
+	}{
+		ObjectReference: objectGlobal.String(),
+		State:           object.Missing,
+	})
+
+	typedChecker := server.PublisherMock.SetTypedChecker(ctx, mc, server)
+	typedChecker.VStateRequest.Set(func(request *payload.VStateRequest) bool {
+		assert.Equal(t, request.Object, objectGlobal)
+		return true
```
sexyhamster, line 205: you can manually create expected VStateReport, and send it with server.SendPayload

ledger-core/virtual/execute/execute.go
AndreyBronin: "object is not ready" means it will be ready later. 
Can we guarantee here that "object does not exist" ?

ledger-core/virtual/integration/method_test.go
AndreyBronin: I am testing stepWaitObjectReady behavior and how it returns correct error

sexyhamster: well you don't create object, you test execute behavior. So VStateRequest/VStateReport is outside of your scope and tested elsewhere

AndreyBronin:
> please, update fields steps, Preconditions, Expected Result in [case 4974](https://insolar.testrail.io/index.php?/cases/view/4974&group_by=cases:section_id&group_order=asc&group_id=917)

done

ledger-core/virtual/execute/execute.go
```diff
@@ -235,13 +235,14 @@ func (s *SMExecute) stepWaitObjectReady(ctx smachine.ExecutionContext) smachine.
 			panic(throw.NotImplemented())
 		}
 	} else if objectState != object.HasState {
-		panic(throw.E("no state on object after readyToWork", struct {
+		s.prepareExecutionError(throw.E("object is not ready", struct {
```
ita-sammann, line 238: I don't like the error message. It doesn't make a lot of sense to the user. Maybe something like "object not found" or "object does not exist" is better?

AndreyBronin: @sexyhamster what do you think about this error message?

sexyhamster: "object does not exist" sounds OK, maybe later on there will be another case for deactivated object

ruz: I would disappoint you and say that this is not correct place to check for anything but Missing. For example if in this point of code the state is Inactive then it doesn't mean that request should fail. The request can be a duplicate and should be finished. Previous panic is not correct and should have been panic(NotImplemented). I insist on returning this new particular error only when `objectState == object.Missing`

AndreyBronin: @ruz  is this correct ? https://github.com/insolar/assured-ledger/pull/400/commits/07eb4c04a21d51116f982474e30b2524db5b5053

ruz: yep, way better. `make generate`

just in case people wonder

- Empty - may happen on recursive call while constructor is running
- Inactive - may happen in this place for two reasons: call on inactive object and duplicate of old request

PR #404, opened 2020-06-26
PLAT-537: integration tests for handler of VFindCallRequest

ruz:
* send VFindCallRequest in variouse situations
* wait for VFindCallResponse
* check response


ruz:
> I like idea of constructing test from smaller parts, but not sure about implementation:
> 
> 1. merge all init functions (init*, it's unneeded to know from test how preparation was done if you can simply look into 1 function, describing all steps is unnecessary)

don't see value in this. I like the way it is.

> 2. s.haveFindResponseSignal isn't closed and no one wrote anything into it

fixed, was a bug

> 3. a lot of usage of mutex, but it's completely unecessary

deleted mutex, left to author of PLAT-389, hope he would run tests with -race

> 4. remake that "construction test from blocks" from string matching (any1 can make a mistake, it's harder to do with functions) to providing functions `type TestComponent func(t *testing.T, suite *vFindCallRequestHandlingTest)` (alternatively it can be method on a "suite")

I don't understand this point.

> 5. no reason to make `vFindCallRequestHandlingTest` to be package local, i'd prefer it to be `VFindCallRequestHandlingSuite` (since it can't be accessed from any other file) and to be related to `VFindCallRequest` and also have `Suite` suffix for it be distinguished from other helper structs

renamed

> 6. also i see a lot of common parts for `vFindCallRequestHandlingTest` and `deduplicateMethodUsingPrevVETest`, i guess when the third test will come - it's time to unify that `Suite` practice

will see

> 7. suite.finish() must be splitted into two separate parts: the first one stops server and MUST be called anyways (in defer, preferably), the second one is `mc.Finish()`/`go leakchecking` and it must be checked in the end of test and only if it passes successfully

done with splitting, not planning to add leak checking in any integration tests, going to create a task

bigbes:
sorry, closed by mistake

PR #408, opened 2020-06-29
PLAT-573: fix pulseNumber in outgoing step; update integration test

ElSacramento:
**About test:**
Change pulse twice: before VCallRequest [B.Bar] and after 
Check pulse and token in VCallRequests + RepeatedCall flag
Set different approvers for delegation tokens + check them
Check that VStateReport isn't sent twice


ledger-core/virtual/integration/outgoing_test.go
```diff
@@ -28,7 +29,21 @@ import (
...
+VCallRequest [B.Bar] + first token
+change pulse -> third pulse
+NO VStateReport
+VDelegatedCallRequest [A] + first token
+VCallRequest [B.Bar] + second token
+VCallResult [A.Foo] + second token
+VDelegatedRequestFinished [A] + second token
+*/
+func TestVirtual_CallMethodOutgoing_WithTwicePulseChange(t *testing.T) {
 	t.Log("C5141")
```
ElSacramento, line 47: I'm gonna fix the case in testrail after merge

ledger-core/virtual/integration/outgoing_test.go
```diff
@@ -105,52 +158,69 @@ func TestVirtual_CallMethodOutgoing_WithPulseChange(t *testing.T) {
...
+					Outgoing:          request.CallOutgoing,
+					DelegateTo:        server.JetCoordinatorMock.Me(),
+				},
+			}
 
-			server.SendPayload(ctx, &payload.VDelegatedCallResponse{
-				Callee:                 request.Callee,
-				ResponseDelegationSpec: expectedToken,
-			})
+			if vdelegatedcallrequestCount == 0 {
```
bigbes, line 183: I'm not sure that is good practice:
1. you can rename that to "firstVDelegatedCallRequest" and use it bool if we should separate first and other message 
2. for only two times you can increment `VDelegatedCallRequestCount` every time and make switch for three cases 0, 1 and default (will lead to `t.Fatal`)

Also, pls, use SnakeCase for names of variables

ledger-core/virtual/integration/outgoing_test.go
```diff
@@ -105,52 +158,69 @@ func TestVirtual_CallMethodOutgoing_WithPulseChange(t *testing.T) {
...
+
+			if vcallrequestCount == 0 {
+				assert.Equal(t, secondPulse, request.CallOutgoing.Pulse()) // new pulse
+				assert.Equal(t, firstExpectedToken, request.DelegationSpec)
+				assert.NotEqual(t, payload.RepeatedCall, request.CallRequestFlags.GetRepeatedCall())
+
 				server.IncrementPulseAndWaitIdle(ctx)
-				callCount++
+				thirdPulse = server.GetPulse().PulseNumber
+				require.Greater(t, thirdPulse.AsUint32(), secondPulse.AsUint32())
```
bigbes, line 210: pls remove that check

ElSacramento: done

ledger-core/virtual/integration/outgoing_test.go
ElSacramento: done

ledger-core/virtual/integration/outgoing_test.go
```diff
@@ -80,10 +131,12 @@ func TestVirtual_CallMethodOutgoing_WithPulseChange(t *testing.T) {
...
 		objectAExecutionMock := runnerMock.AddExecutionMock("Foo")
 		objectAExecutionMock.AddStart(
-			func(ctx execution.Context) {
+			func(_ execution.Context) {
 				logger.Debug("ExecutionStart [A.Foo]")
-				assert.Equal(t, objectAGlobal, ctx.Object)
-				assert.Equal(t, callOutgoing, ctx.Request.CallOutgoing)
+				server.IncrementPulseAndWaitIdle(ctx)
+				secondPulse = server.GetPulse().PulseNumber
+				require.Greater(t, secondPulse.AsUint32(), firstPulse.AsUint32())
```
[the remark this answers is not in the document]
ElSacramento: done

ledger-core/virtual/integration/outgoing_test.go
MariaVasilenko: why you can't done this now?

MariaVasilenko: now we have simple process - reviewer see code, logic and testrail case. i have question about this changes - test with one pulse change(testrail old case) is not need? why?

ElSacramento: because the case is already passed in testrail, I don't want to make a misunderstanding while the branch isn't merged.

ElSacramento: test with two pulse change contains one pulse change, so that test with one pulse change will duplicate this test.  
if the problem occurs in one pulse change we will have two fails, I do think that it's better to have one test.

ElSacramento: done

ledger-core/virtual/integration/outgoing_test.go
```diff
@@ -105,52 +157,68 @@ func TestVirtual_CallMethodOutgoing_WithPulseChange(t *testing.T) {
...
 				// omit sending
 				return false
 			}
 
-			assert.Equal(t, payload.RepeatedCall, request.CallRequestFlags.GetRepeatedCall())
+			assert.Equal(t, secondPulse, request.CallOutgoing.Pulse()) // the same pulse
 
-			assert.Equal(t, expectedToken, request.DelegationSpec)
+			// reRequest -> check all fields
+			expectedVCallRequest.CallOutgoing = request.CallOutgoing
```
MariaVasilenko, line 220: why you disabled field from check?

ElSacramento: I can only check pulse and I've done it above.

PR #414, opened 2020-06-30
NOISSUE: sync pending table realization with working table

sexyhamster:
**- What I did**
- sync tables behaviour
- hide list method that should not be called outside of table
- added bench for memory usage

before
```
pkg: github.com/insolar/assured-ledger/ledger-core/virtual/object
BenchmarkPendingTable-12    	 3951997	       285 ns/op	     304 B/op	       6 allocs/op
BenchmarkWorkingTable-12    	 4537809	       244 ns/op	     240 B/op	       6 allocs/op
```
after
```
goos: darwin
goarch: amd64
pkg: github.com/insolar/assured-ledger/ledger-core/virtual/object
BenchmarkPendingTable-12    	 7344967	       154 ns/op	     160 B/op	       4 allocs/op
BenchmarkWorkingTable-12    	 5528679	       192 ns/op	     208 B/op	       5 allocs/op
```

**- How I did it**

**- How to verify it**

### DIFF
--- a/ledger-core/virtual/object/request_table.go
+++ b/ledger-core/virtual/object/request_table.go
@@ -13,20 +13,22 @@ import (
 )
 
 type PendingTable struct {
-	lists map[contract.InterferenceFlag]*PendingList
+	lists []*PendingList
 }
 
 func NewRequestTable() PendingTable {
 	var rt PendingTable
-	rt.lists = make(map[contract.InterferenceFlag]*PendingList)
 
-	rt.lists[contract.CallTolerable] = NewRequestList()
-	rt.lists[contract.CallIntolerable] = NewRequestList()
+	rt.lists = make([]*PendingList, contract.InterferenceFlagCount)
+	for i := 1; i < contract.InterferenceFlagCount; i++ {
+		rt.lists[i] = NewRequestList()
+	}
+
 	return rt
 }
 
 func (rt *PendingTable) GetList(flag contract.InterferenceFlag) *PendingList {
-	if flag.IsZero() {
+	if flag.IsZero() || flag >= contract.InterferenceFlagCount {
 		panic(throw.IllegalValue())
 	}
 	return rt.lists[flag]
--- a/ledger-core/virtual/object/request_table.go
+++ b/ledger-core/virtual/object/request_table.go
@@ -13,13 +13,12 @@ import (
 )
 
 type PendingTable struct {
-	lists []*PendingList
+	lists [3]*PendingList
 }
 
 func NewRequestTable() PendingTable {
 	var rt PendingTable
 
-	rt.lists = make([]*PendingList, contract.InterferenceFlagCount)
 	for i := 1; i < contract.InterferenceFlagCount; i++ {
 		rt.lists[i] = NewRequestList()
 	}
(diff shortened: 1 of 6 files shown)


ledger-core/virtual/object/working_table.go
```diff
@@ -45,11 +43,11 @@ func (wt *WorkingTable) GetResults() map[reference.Global]Summary {
 }
 
 func (wt *WorkingTable) Add(flag contract.InterferenceFlag, ref reference.Global) bool {
```
cyraxred, line 45: Semantically "Put" fits better here than "Add"

ledger-core/virtual/object/request_table.go
```diff
@@ -13,20 +13,22 @@ import (
 )
 
 type PendingTable struct {
-	lists map[contract.InterferenceFlag]*PendingList
+	lists []*PendingList
```
cyraxred, line 16: there is even no need for a slice - better to use an array

ledger-core/virtual/object/working_table.go
sexyhamster: From my experience - Put is replace, while add is not replace. Maybe that comes from REST :) I'd prefer leave it as it is

ledger-core/virtual/object/request_table.go
sexyhamster: good point, will fix:
```
goos: darwin
goarch: amd64
pkg: github.com/insolar/assured-ledger/ledger-core/virtual/object
BenchmarkPendingTable-12    	 7344967	       154 ns/op	     160 B/op	       4 allocs/op
BenchmarkWorkingTable-12    	 5528679	       192 ns/op	     208 B/op	       5 allocs/op
```

ledger-core/virtual/object/request_table.go
```diff
@@ -13,27 +13,27 @@ import (
...
 	}
 
 	return rt
 }
 
-func (rt *PendingTable) GetList(flag contract.InterferenceFlag) *PendingList {
-	if flag.IsZero() || flag >= contract.InterferenceFlagCount {
-		panic(throw.IllegalValue())
+func (rt PendingTable) GetList(flag contract.InterferenceFlag) *PendingList {
+	if flag > 0 && int(flag) <= len(rt.lists) {
```
sexyhamster, line 30: this one is harder to read, from my point of view.
I prefer 
```
if check for bad value {
  handle bad value
}
now we have good value
```

ledger-core/virtual/object/working_table.go
```diff
@@ -14,38 +14,38 @@ import (
...
 	return wt.results
 }
 
-func (wt *WorkingTable) Add(flag contract.InterferenceFlag, ref reference.Global) bool {
+func (wt WorkingTable) Add(flag contract.InterferenceFlag, ref reference.Global) bool {
 	return wt.GetList(flag).add(ref)
 }
 
-func (wt *WorkingTable) SetActive(flag contract.InterferenceFlag, ref reference.Global) bool {
+func (wt WorkingTable) SetActive(flag contract.InterferenceFlag, ref reference.Global) bool {
```
sexyhamster, line 48: this is not really good, because it relies on map shallow copy mechanics. If the object is changed by method, it's better to actually have ptr receiver

PR #417, opened 2020-06-30
PLAT-558: CallOutgoing fix

sexyhamster:
**- What I did**
- switch to proper CallOutgoing, now it is affined with caller
- Fixed all the tests

**- How to verify it**
run tests


ledger-core/insolar/payload/marker_auth.go
```diff
@@ -38,7 +40,15 @@ func (m *VStateRequest) customSubject() (Reference, authSubjectMode) {
 }
 
 func (m *VCallResult) customSubject() (Reference, authSubjectMode) {
-	return m.GetCallee(), hasAuthSubject
+	switch m.CallType {
```
sexyhamster, line 43: Now I proxy Callee from VCallRequest as is. 

I'm not sure about this behaviour, stop me if I'm wrong
@ivanshibitov 
@cyraxred

ledger-core/virtual/execute/execute.go
```diff
@@ -109,13 +109,13 @@ func (s *SMExecute) prepareExecution(ctx context.Context) {
...
 		s.isConstructor = true
-		s.execution.Object = reference.NewSelf(s.Payload.CallOutgoing)
+		s.execution.Object = s.Payload.CallOutgoing
 	} else {
 		s.execution.Object = s.Payload.Callee
 	}
 
-	s.execution.Incoming = reference.NewRecordOf(s.Payload.Caller, s.Payload.CallOutgoing)
-	s.execution.Outgoing = reference.NewRecordOf(s.Payload.Callee, s.Payload.CallOutgoing)
+	s.execution.Incoming = reference.NewRecordOf(s.Payload.Callee, s.Payload.CallOutgoing.GetLocal())
```
sexyhamster, line 117: so now Incoming is affined to callee, outgoing should be affined with caller, and now is based on VCallRequest

ledger-core/virtual/execute/execute_semiintegration_test.go
```diff
@@ -428,7 +423,7 @@ func TestSMExecute_Semi_MethodOnEmptyObject(t *testing.T) {
 			CallFlags:    payload.BuildCallFlags(contract.CallTolerable, contract.CallDirty),
 			CallOutgoing: outgoing,
 
-			Caller:         caller,
+			Caller:         outgoing,
```
ElSacramento, line 426: Caller and CallOutgoing are the same? It's not a typo, is it?

sexyhamster: typo, will check, but it doesnt matter in this test case

ledger-core/insolar/payload/marker_auth.go
```diff
@@ -38,7 +40,15 @@ func (m *VStateRequest) customSubject() (Reference, authSubjectMode) {
...
+	case CTMethod:
+		return m.GetCallee(), hasAuthSubject
+	case CTConstructor:
+		return m.GetCallOutgoing(), hasAuthSubject
```
cyraxred, line 47: this looks weird to me ... why Callee is different for constructor?

ledger-core/virtual/execute/execute.go
cyraxred: Ok for R0

ledger-core/virtual/execute/execute_publish_callsummary_test.go
```diff
@@ -33,21 +34,20 @@ func TestSMExecute_PublishVCallResultToCallSummarySM(t *testing.T) {
 
 		pd          = pulse.NewFirstPulsarData(10, longbits.Bits256{})
 		pulseSlot   = conveyor.NewPresentPulseSlot(nil, pd.AsRange())
-		smObjectID  = gen.UniqueLocalRefWithPulse(pd.PulseNumber)
-		smGlobalRef = reference.NewSelf(smObjectID)
+		outgoingRef = reference.NewRecordOf(gen.UniqueGlobalRef(), gen.UniqueLocalRefWithPulse(pd.PulseNumber))
```
cyraxred, line 37: well would be a bit better to ensure base.PN() <= Local.PN() - here and anywhere else

ledger-core/insolar/payload/marker_auth.go
sexyhamster: removed

ledger-core/insolar/payload/marker_auth.go
sexyhamster: removed

ledger-core/virtual/execute/execute_semiintegration_test.go
sexyhamster: fixed

ledger-core/virtual/execute/execute_publish_callsummary_test.go
sexyhamster: Until it is not mandatory, I'm not sure that this makes sense. Every new test need to make sure that callee and caller must be from past PN, and sometimes test does not care about PN at all. I prefere to do this kind of fix with mandatory constraint for NewRecordOf, so all the cases will be covered

ledger-core/application/testwalletapi/statemachine/statemachine.go
```diff
@@ -106,7 +105,7 @@ func (s *SMTestAPICall) stepSendRequest(ctx smachine.ExecutionContext) smachine.
 	case payload.CTMethod:
 		obj = s.requestPayload.Callee
 	case payload.CTConstructor:
-		obj = reference.NewSelf(s.requestPayload.CallOutgoing)
+		obj = s.requestPayload.CallOutgoing
```
ruz, line 108: Strange that nobody questioned this change :(

PR #418, opened 2020-07-01
NOISSUE: Helper for minimock to report full stacktrace

cyraxred:
**- What I did**
* mocklog.T(t) that usable for minimocks to report full stacktrace on errors/fatals


ledger-core/vanilla/args/var_agrs.go
```diff
@@ -0,0 +1,17 @@
+// Copyright 2020 Insolar Network Ltd.
```
ruz, line 1: file name s/var_agrs/var_args/

PR #420, opened 2020-07-01
PLAT-488: A.New->B.Bar, twice pulse change

ElSacramento:
**- What I did**

**- How I did it**

**- How to verify it**


ledger-core/virtual/integration/outgoing_test.go
```diff
@@ -266,3 +266,234 @@ func TestVirtual_CallMethodOutgoing_WithTwicePulseChange(t *testing.T) {
...
+			assert.Equal(t, payload.Empty, report.Status)
+			assert.Zero(t, report.DelegationSpec)
+			return false
+		})
+		typedChecker.VDelegatedCallRequest.Set(func(request *payload.VDelegatedCallRequest) bool {
+			assert.Equal(t, outgoing, request.Callee)
+
+			msg := payload.VDelegatedCallResponse{
+				Callee: request.Callee,
+				ResponseDelegationSpec: payload.CallDelegationToken{
```
sexyhamster, line 403: here you are actually building tokens. I think you can fill firstExpectedToken/secondExpectedToken here with all the fields.

ElSacramento: I just want to use different variables to be sure that they don't affect each other.

sexyhamster: well, system logic is - token that is provided by VDelegatedCallResponse should be used in messages. So they are supposed to be the same. And creating them here and sending them back to node should not be a problem (given that VDelegatedCallRequest/Response have tests)

ElSacramento: done

PR #424, opened 2020-07-01
PLAT-522, PLAT-521, PLAT-586: add test for double pulse change

MariaVasilenko:
- joined 3 cases to table test(jira PLAT-522,PLAT-521, PLAT-586)
- updated the [case](https://insolar.testrail.io/index.php?/cases/view/5211&group_by=cases:section_id&group_order=asc&group_id=946) 
- @bigbes fixed VDelegatedRequestFinished.LatestState for unordered calls
- stepWaitExecutionResult is public


ledger-core/testutils/predicate/predicate.go
```diff
@@ -53,3 +53,22 @@ func Or(predicates ...Func) Func {
 		return false
 	}
 }
+
+func Sequence(predicates ...Func) Func {
```
bigbes, line 57:
```suggestion
// ChainOf will wait for all predicates to fire on different steps
// for example:
// ChainOf(X1, X2, X3) will wait for:
// * predicate X1 eventually to return true
// * predicate X2 eventually to return true (no more checks of X1 ever)
// * and predicate X3 eventually to return true (no checks of X1 and X2)
// and only then it'll return true
func ChainOf(predicates ...Func) Func {```

MariaVasilenko: ops( sorry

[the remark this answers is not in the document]
MariaVasilenko: done

PR #426, opened 2020-07-02
PLAT-591: proper suite finalization

sexyhamster:
**- What I did**
Fixed leaking SMs after test is finished:

1. we need to ensure all async calls are done
2. We need to wait that all sm finished work

### DIFF
--- a/ledger-core/virtual/integration/deduplication/find_call_request_handler_test.go
+++ b/ledger-core/virtual/integration/deduplication/find_call_request_handler_test.go
@@ -177,7 +177,13 @@ func TestDeduplication_VFindCallRequestHandling(t *testing.T) {
 
 			suite.waitFindRequestResponse(t)
 
-			suite.finish(t)
+			if suite.finalizedMessageSent != nil {
+				testutils.WaitSignalsTimed(t, 10*time.Second, suite.finalizedMessageSent)
+			}
+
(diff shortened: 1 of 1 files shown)


ledger-core/virtual/integration/deduplication/find_call_request_handler_test.go
```diff
@@ -482,7 +482,9 @@ func (s *VFindCallRequestHandlingSuite) addPayloadAndWaitIdle(
 	s.server.WaitActiveThenIdleConveyor()
 }
 
-func (s *VFindCallRequestHandlingSuite) finish() {
+func (s *VFindCallRequestHandlingSuite) finish(t *testing.T) {
+	testutils.WaitSignalsTimed(t, 10*time.Second, s.server.Journal.WaitAllAsyncCallsDone())
+	s.server.WaitIdleConveyor()
```
bigbes, line 487: Why that line is needed?

sexyhamster: because we need to finish all the sms that were waiting for async calls to finish their jobs

sexyhamster: switched to waiting of StateReport message

PR #427, opened 2020-07-02
PLAT-575: skip check auth of some messages

ivanshibitov:
**- What I did**
Add support of messages which can come from any pulse

**- How I did it**

**- How to verify it**

### DIFF
--- a/ledger-core/virtual/authentication/service.go
+++ b/ledger-core/virtual/authentication/service.go
@@ -90,11 +90,11 @@ func (s service) getExpectedVE(ctx context.Context, subjectRef reference.Global,
 
 func (s service) IsMessageFromVirtualLegitimate(ctx context.Context, payloadObj interface{}, sender reference.Global, pr pulse.Range) (bool, error) {
 	verifyForPulse := pr.RightBoundData().PulseNumber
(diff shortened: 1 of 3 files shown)


ledger-core/virtual/authentication/service.go
```diff
@@ -90,11 +90,11 @@ func (s service) getExpectedVE(ctx context.Context, subjectRef reference.Global,
 
 func (s service) IsMessageFromVirtualLegitimate(ctx context.Context, payloadObj interface{}, sender reference.Global, pr pulse.Range) (bool, error) {
 	verifyForPulse := pr.RightBoundData().PulseNumber
-	subjectRef, usePrev, ok := payload.GetSenderAuthenticationSubjectAndPulse(payloadObj)
+	subjectRef, mode, ok := payload.GetSenderAuthenticationSubjectAndPulse(payloadObj)
```
sexyhamster, line 93: I'd do this more like 
```
if !ok {
   return false, throw.New("Unexpected message type")
}

switch mode {
    case payload.UsePrevPulse:
    case payload.UseAnyPulse:
    default:
         panic(throw.IllegalValue())
}
```

But it's ok like this too

ivanshibitov: will do in next pr

PR #428, opened 2020-07-02
PLAT-578: disable graceful stop for network. Will fix it later on

sexyhamster:
**- What I did**
commented graceful stop logic. it is broken and we need to fix it in R1 with new network part

**- How I did it**

**- How to verify it**
Run make test_func, make sure that no insolard nodes left working


AndreyBronin:
May be better to comment 	`n.TerminationHandler.Leave(ctx, 0)` inside `func (n *ServiceNetwork) GracefulStop(ctx context.Context) error {}` ??

sexyhamster:
Good point. It will look better :)

PR #432, opened 2020-07-02
Plat-536: fix process message in past pulse

ivanshibitov:
**- What I did**
stop execute if we are not in present slot

**- How I did it**

**- How to verify it**


ledger-core/virtual/handlers/vcallrequest.go
```diff
@@ -53,6 +54,10 @@ func (s *SMVCallRequest) Init(ctx smachine.InitializationContext) smachine.State
 }
 
 func (s *SMVCallRequest) stepExecute(ctx smachine.ExecutionContext) smachine.StateUpdate {
+	if s.pulseSlot.State() != conveyor.Present {
```
bigbes, line 57: should we duplicate that check here and [there](https://github.com/insolar/assured-ledger/pull/432/files#diff-b9bd8941b93f6eedb3ea9c2e476f935dR132-R135)?

Maybe we can only check in SMExecute?

ledger-core/virtual/handlers/handle_meta.go
```diff
@@ -62,6 +70,7 @@ func (f FactoryMeta) Process(msg *statemachine.DispatcherMessage, pr pulse.Range
 
 	logger.Info(logProcessing{
 		messageType: fmt.Sprintf("id=%d, type=%s", payloadTypeID, payloadType.String()),
+		pulseData:   pr,
```
sexyhamster, line 73: some extra debug?

ivanshibitov: actually, don't need it

ledger-core/virtual/execute/execute.go
```diff
@@ -131,6 +131,10 @@ func (s *SMExecute) migrationDefault(ctx smachine.MigrationContext) smachine.Sta
 }
 
 func (s *SMExecute) Init(ctx smachine.InitializationContext) smachine.StateUpdate {
+	if s.pulseSlot.State() != conveyor.Present {
```
bigbes, line 134: No need to check it here, SMExecute isn't handler and can only be spawned in handler :)

PR #436, opened 2020-07-03
PLAT-553: retries limit in SMTestAPICall and fixes

ruz:
* limit number of retries
* object was not calculated correctly constructors
* idea of unpublishing bargein was not working, pulse was the same all
  the time, just placed first send into one step and resend into
  another
* set migration after first send
* write tests

### DIFF
--- a/ledger-core/application/testwalletapi/statemachine/statemachine_test.go
+++ b/ledger-core/application/testwalletapi/statemachine/statemachine_test.go
@@ -14,9 +14,6 @@ import (
 	"github.com/stretchr/testify/require"
 	"gotest.tools/assert"
 
-	"github.com/insolar/assured-ledger/ledger-core/application/builtin/proxy/testwallet"
-	"github.com/insolar/assured-ledger/ledger-core/conveyor/smachine"
-	"github.com/insolar/assured-ledger/ledger-core/insolar"
 	"github.com/insolar/assured-ledger/ledger-core/insolar/contract"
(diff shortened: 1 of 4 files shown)


ledger-core/application/testwalletapi/statemachine/statemachine_test.go
```diff
@@ -118,43 +124,72 @@ func TestSMTestAPICall_Migrate_After_SendRequest(t *testing.T) {
...
+
+	smWrapper := slotMachine.AddStateMachine(ctx, &smRequest)
+
+	messageSent := make(chan struct{}, 1)
+	slotMachine.MessageSender.SendRole.Set(func(_ context.Context, msg payload.Marshaler, role node.DynamicRole, object reference.Global, pn pulse.Number, _ ...messagesender.SendOption) error {
+		messageSent <- struct{}{}
+		return nil
+	})
+
+	for i := 0; i < 3; i++ {
```
ivanshibitov, line 183: may be replace '3' -> 'MaxRepeats' contant

ruz: nope, this is a test, change of number of max retries in code should result in test failure, I will add a constant

ruz: added a constant

ivanshibitov: hmmmm. okay!

PR #444, opened 2020-07-06
[Merged by Bors] - PLAT-616: fix test TestVirtual_Constructor_PulseChangedWhileOutgoing

ruz:
* wait for message to be send before asserting counter
* test case expected execution to be finished and only after that
  VStateRequest should be send into the system

### DIFF
--- a/ledger-core/virtual/integration/mock/typepublishchecker.go
+++ b/ledger-core/virtual/integration/mock/typepublishchecker.go
@@ -7,7 +7,7 @@ package mock
 
 import (
 	"context"
-	mm_time "time"
+	"time"
 
 	"github.com/ThreeDotsLabs/watermill/message"
 	"github.com/gojuno/minimock/v3"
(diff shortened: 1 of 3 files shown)


ledger-core/virtual/integration/mock/typepublishchecker.go
```diff
@@ -634,3 +671,19 @@ func (p *TypePublishChecker) MinimockWait(timeout mm_time.Duration) {
 		}
 	}
 }
+
+func waitCounterIndefinitely(counter *atomickit.Int, count int) synckit.SignalChannel {
```
bigbes, line 675: There should be mechanism (even not "thread safe") to stop waiting, in case we stop test with timeout.

ruz: I couldn't find good solution to implement such mechanizm and decided that it's ok. Do you have simple and straightforward solution?

ruz: do we really expect test failures to cleanup everything after?

PR #447, opened 2020-07-06
NOISSUE: (R1) drop storage

cyraxred:
**- What I did**

**- How I did it**

**- How to verify it**


ledger-core/ledger/server/buildersvc/update_future.go
```diff
@@ -6,18 +6,79 @@
...
+	switch {
+	case err != nil:
+		if len(allocations) != 0 {
+			panic(throw.IllegalValue())
+		}
+	case len(allocations) == 0:
+		panic(throw.IllegalValue())
+	}
+
+	p.mutex.Lock()
```
sexyhamster, line 74: this could be a method like
```
func (p *Future) setAllocations(allocations []ledger.DirectoryIndex, err error) bool {
   p.mutex.Lock()
   defer p.mutex.Unlock()
   if p.err != nil || p.allocations != nil {
      return false
   }
   p.allocations, p.err = allocations, err
   return true
}
```

sexyhamster: it would be inline, so it won't affect performance, but it easier to ensure that mutex will be unlocked in all the branches

ledger-core/ledger/server/datawriter/smplash.go
```diff
@@ -86,13 +89,19 @@ func (p *SMStreamDropBuilder) stepInit(ctx smachine.InitializationContext) smach
 // 	return prevPN
 // }
 
-func (p *SMStreamDropBuilder) stepCreateStreamDrop(ctx smachine.ExecutionContext) smachine.StateUpdate {
+func (p *SMPlash) stepCreatePlush(ctx smachine.ExecutionContext) smachine.StateUpdate {
+
+	var (
+		tree jet.Tree
+		pop census.OnlinePopulation
```
sexyhamster, line 96: I think that this is too tightly coupled with census. There should be something like API between those, but it is already "gcpv2/api/census", it looks like those layers are mixed up

ledger-core/virtual/execute/execute.go
```diff
@@ -9,7 +9,6 @@ package execute
 
 import (
 	"context"
-
```
sexyhamster, line 12: this is probably not intentional

ledger-core/ledger/server/datawriter/smplash.go
cyraxred: census is separate from gcpv2 - it was only created as a part of it and can be moved out. I don't touch its location just for sake of smaller PR

ledger-core/virtual/execute/execute.go
cyraxred: definitely it is not ... some random tap?

ledger-core/ledger/server/buildersvc/update_future.go
cyraxred: fixed

ledger-core/virtual/execute/execute.go
cyraxred: fixed

ledger-core/conveyor/smachine/exec_adapter.go
```diff
@@ -33,20 +33,41 @@ func (p ExecutionAdapter) GetAdapterID() AdapterID {
...
 func (p ExecutionAdapter) PrepareNotify(ctx ExecutionContext, fn AdapterNotifyFunc) NotifyRequester {
+	if fn == nil {
+		panic(throw.IllegalValue())
+	}
 	ec := ctx.(*executionContext)
 	return &adapterNotifyRequest{ctx: ec, fn: fn, adapterID: p.adapterID, isLogging: ec.s.getAdapterLogging(),
 		executor: p.executor, mode: adapterAsyncCallContext}
 }
+
+func (p ExecutionAdapter) SendFailureNotify(ctx FailureContext, fn AdapterNotifyFunc) {
```
ruz, line 63: split it in two: PrepareFailureNotify and SendFailureNotify. the last one is syntax sugar on top of first. May be I'm wrong and DelayedSend doesn't make sense for FailureNotify.

ledger-core/conveyor/smachine/slot_machine.go
```diff
@@ -1100,12 +1100,39 @@ func (m *SlotMachine) activateDependantByDetachable(links []StepLink, ignore Slo
 		return true
 	}
 
-	m.syncQueue.AddAsyncUpdate(SlotLink{}, func(_ SlotLink, worker FixedSlotWorker) {
+	if !m.syncQueue.AddAsyncUpdate(SlotLink{}, func(_ SlotLink, worker FixedSlotWorker) {
```
ruz, line 1103: I don't insist, but I would like the following code instead of this:
`
added := m.syncQueue.AddAsyncUpdate(...)
if !added {
    ...
}
`
It's just way easier to read

ledger-core/ledger/api_section.go
```diff
@@ -0,0 +1,102 @@
...
+	return Ordinal(v)
+}
+
+type ChapterID uint32
+const MaxChapterID ChapterID = 0x00FF_FFFF
+const MaxChapterOffset = 0x00FF_FFFF
+
+type StorageLocator uint64
+
+func NewLocator(id SectionID, chapterID ChapterID, ofs uint32) StorageLocator {
```
ruz, line 64: s/ofs/offset/ please

ledger-core/conveyor/smachine/exec_adapter.go
cyraxred: DelayedSend can't be applied here, as FailureContext expects no result, hence there will be only one method - Send()

ledger-core/conveyor/smachine/slot_machine.go
cyraxred: well, this is a matter of taste - this reads ok to me, and it doesn't pollute namespace, which I found very useful - no need to pollute code with { ... }

ledger-core/ledger/api_section.go
cyraxred: any other meaning for "ofs"? ;)

ledger-core/ledger/server/buildersvc/bundle/writer.go
```diff
@@ -0,0 +1,244 @@
...
+	case ok:
+		defer func() {
+			next <- struct{}{}
+			close(next)
+		}()
+	default:
+		defer close(next)
+	}
+
+	fn(true)
```
ruz, line 76: above block can be written differently:

```
defer close(next)
if ok {
    defer func() { next <- struct{}{} }()
}
if fn != nil {
    fn(true)
}
return true
```

or 

```
defer func() {
  if ok {
    next <- struct{}{}
  }
  close(next)
}()


if fn != nil {
    fn(true)
}
return true
```

Either I find more readable than above switch.

ledger-core/ledger/server/buildersvc/bundle/writer.go
```diff
@@ -0,0 +1,244 @@
...
+	locked := false
+	chained := false
+	rollback := true
+	var err error
+
+	defer func() {
+		switch {
+		case !rollback:
+			if locked {
+				p.mutex.Unlock()
```
ruz, line 161: this is impossible case.

ledger-core/ledger/server/buildersvc/bundle/writer.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	prev, next := p._prepareWait(true)
+	go p.applyBundleSafe(snapshot, bundle, prev, next, completedFn)
+
+	snapshot = nil
+	return nil
+}
+
+func (p *bundleWriter) applyBundleSafe(snapshot Snapshot, bundle Writeable,
```
ruz, line 147: this function way too complex

ledger-core/ledger/server/buildersvc/bundle/writer.go
cyraxred: it is, but this is a precaution for someone to try to touch this code

ledger-core/ledger/server/buildersvc/bundle/writer.go
cyraxred: it does only one thing - proper ordering. Due to sensitivity it was put in one place in one function. Do you want this complexity to be spread across more code and to be more prone to error?

ledger-core/ledger/server/buildersvc/bundle/writer.go
cyraxred: fine by me :)

ledger-core/ledger/server/buildersvc/bundle/writer.go
cyraxred: will write more explanatory text here

ledger-core/ledger/server/buildersvc/bundle/writer.go
cyraxred: but reworked it for clarity

PR #453, opened 2020-07-06
[Merged by Bors] - PLAT-618: incoming call when object in pending constructor state

ruz:


ruz:
bors merge

ledger-core/virtual/integration/method_pulse_change_test.go
```diff
@@ -517,15 +514,7 @@ func TestVirtual_MethodCall_IfConstructorIsPending(t *testing.T) {
...
-				typedChecker.VDelegatedRequestFinished.Set(func(res *payload.VDelegatedRequestFinished) bool {
-					getDelegated = true
-					require.Equal(t, delegationToken, res.DelegationSpec)
-					return false
-				})
-				typedChecker.VDelegatedCallResponse.Set(func(res *payload.VDelegatedCallResponse) bool {
-					delegationToken = res.ResponseDelegationSpec
-					return false
-				})
+				typedChecker.VDelegatedCallResponse.SetResend(false)
```
MariaVasilenko, line 517: then add check VDelegatedCallResponse count before mc.Finish()

ruz: why? we test logic around delegations a lot. I think that in this scenario it's most important to check memory in executor mock and message with result of execution. That's it and we do both.

ledger-core/virtual/object/object.go
```diff
@@ -42,6 +42,8 @@ const (
 
 const waitStatePulsePercent = 10
 
+const UnorderedMaxParallelism = 30
```
ita-sammann, line 45: Is it a temporary limit until PLAT-429 (hierarchical semaphores) is done?

ruz: this is what we had in the code, I just moved it into constant.

PR #458, opened 2020-07-07
[Merged by Bors] - Plat-570: test dt check fail if no dt and source != expectedVE

MikhailZhukov:
Test that execution will stopped if message have zero DT and source not equal expected VE

Combine test for PLAT-570, PLAT-565 and PLAT-564 into one tabular test

**- What I did**

**- How I did it**

**- How to verify it**


ita-sammann:
bors ping

PR #468, opened 2020-07-08
PLAT-367: get rid of IgnoreAllErrors from most of the tests

sexyhamster:
The only place that still holds IgnoreError are network integration tests, which would be reworked

### DIFF
--- a/ledger-core/application/api/availability_test.go
+++ b/ledger-core/application/api/availability_test.go
@@ -43,7 +43,8 @@ func TestAvailabilityChecker_UpdateStatus(t *testing.T) {
 	instestlogger.SetTestOutputWithErrorFilter(t, func(s string) bool {
 		expectedError := strings.Contains(s, "no response or bad StatusCode") ||
 			strings.Contains(s, "Can't decode body: invalid character '}'") ||
-			strings.Contains(s, "connection refused")
+			strings.Contains(s, "connection refused") ||
+			strings.Contains(s, "No connection could be made because the target machine actively refused it")
 		return !expectedError
 	})
 	ctx, _ := inslogger.InitNodeLoggerByGlobal("", "")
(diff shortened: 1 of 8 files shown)


ledger-core/application/api/availability_test.go
```diff
@@ -39,7 +40,12 @@ func waitForStatus(t *testing.T, nc *api.NetworkChecker, expected bool) {
 }
 
 func TestAvailabilityChecker_UpdateStatus(t *testing.T) {
-	instestlogger.SetTestOutputWithIgnoreAllErrors(t)
+	instestlogger.SetTestOutputWithErrorFilter(t, func(s string) bool {
+		expectedError := strings.Contains(s, "no response or bad StatusCode") ||
+			strings.Contains(s, "Can't decode body: invalid character '}'") ||
+			strings.Contains(s, "connection refused")
```
aalekseev-ru, line 46: maybe move those strings to constants?

sexyhamster: what do you mean?

sexyhamster: they are not constants. That couple of expected errors that are in this test

ita-sammann:
test

ledger-core/server/internal/headless/components_test.go
```diff
@@ -20,7 +21,12 @@ import (
 )
 
 func TestComponents(t *testing.T) {
-	instestlogger.SetTestOutputWithIgnoreAllErrors(t)
+	instestlogger.SetTestOutputWithErrorFilter(t, func(s string) bool {
+		if strings.Contains(s, "Failed to export to Prometheus: cannot register the collector: duplicate metrics collector registration attempted") {
```
cyraxred, line 25: also happens
"Failed to export to Prometheus: cannot register the collector: descriptor Desc ..."
so I suggest to use 
"Failed to export to Prometheus: cannot register the collector"

PR #472, opened 2020-07-09
[Merged by Bors] - Rename test case C4930 to C5094

IgorMarnat:
**- What I did**
There were two duplicated test cases, C4930 and C5094 in the testrail. C4930 had implemented test in the code but the test case was put in the wrong section in testrail. I should have moved it to the right section and removed C5094, but I didn't know that id of the test case cannot be changed and removed C4930 instead. Now it's easier just to change test case id in the code so that it corresponds to the test case id in the testrail (C5094).

**- How I did it**

**- How to verify it**


IgorMarnat:
bors merge

PR #477, opened 2020-07-09
[Merged by Bors] - PLAT-563: correct DT will not fail and SM will be created

MikhailZhukov:
Test that correct delegation token will not fail and SM will be created

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/ledger-core/virtual/integration/delegationtoken_test.go
+++ b/ledger-core/virtual/integration/delegationtoken_test.go
@@ -28,6 +28,89 @@ import (
 	"github.com/insolar/assured-ledger/ledger-core/virtual/testutils"
 )
 
+var messagesWithToken = []struct {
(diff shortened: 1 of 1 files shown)


ledger-core/virtual/integration/delegationtoken_test.go
```diff
@@ -28,6 +28,89 @@ import (
 	"github.com/insolar/assured-ledger/ledger-core/virtual/testutils"
 )
 
+var messagesWithToken = []struct {
```
aalekseev-ru, line 31: it means, they may have tokens?

MikhailZhukov: Yes, this is list  of messages which can have token

PR #491, opened 2020-07-12
NOISSUE: use object reference as publishing key for object's shared data

ruz:
Stringification of reference is way to slow for such purpose

In theory we can wrap reference into special struct, but in this case
I think that this is object's primary data and it's ok to use object's
reference directly.


ruz:
@bigbes, main reason for this change is "Stringification of reference is way to slow for such purpose". I've changed type from `string` to `reference.Global`. Quality of code in respect to "collisions and future bugs" has not changed. Reference as key is similar to how it was before, just different form of the same reference. It's ~48 times faster to access a map using reference directly instead of stringifying first.

I've considered changing this to some struct named "ObjectSharedDataKey", weighted two variants and decided not to change this aspect. I wrote the paragraph in commit message, the one you quote, especially to avoid this conversation.

PR #493, opened 2020-07-13
PLAT-516: parallel call of two ordered methods

IgorMarnat:
Added test TestVirtual_Method_WithoutExecutor_Ordered which checks that ordered methods are executed sequentially, second is blocked until the first one is done.


ledger-core/virtual/integration/method_test.go
```diff
@@ -343,6 +344,90 @@ func TestVirtual_Method_WithoutExecutor_Unordered(t *testing.T) {
...
+			server.SendPayload(ctx, &pl)
+		}
+	}
+	awaitFullStop := server.Journal.WaitStopOf(&execute.SMExecute{}, 2)
+	testutils.WaitSignalsTimed(t, 10*time.Second, awaitFullStop)
+	testutils.WaitSignalsTimed(t, 10*time.Second, server.Journal.WaitAllAsyncCallsDone())
+	{
+		assert.Equal(t, 2, typedChecker.VCallResult.Count())
+
+		if !server.PublisherMock.WaitCount(countBefore+2, 10*time.Second) {
```
bigbes, line 423: that part is not needed

ledger-core/virtual/integration/method_test.go
```diff
@@ -343,6 +344,90 @@ func TestVirtual_Method_WithoutExecutor_Unordered(t *testing.T) {
...
+			})
+			runnerMock.AddExecutionClassify(key, contract.MethodIsolation{
+				Interference: interferenceFlag,
+				State:        stateFlag,
+			}, nil)
+
+			server.SendPayload(ctx, &pl)
+		}
+	}
+	awaitFullStop := server.Journal.WaitStopOf(&execute.SMExecute{}, 2)
```
bigbes, line 417: please move that part to place before messages were sent (as we discussed before)

ledger-core/virtual/integration/method_test.go
IgorMarnat: Fixed

ledger-core/virtual/integration/method_test.go
IgorMarnat: Right. Done.

PR #497, opened 2020-07-13
NOISSUE: Architectural supervision fixes

dgovorukhin:
Fixed all comments from https://docs.google.com/document/d/1LEGQZtAzt-Ii3JOiSlLWyiHR08UaMAjbxp1xmooyIa4/edit#

### DIFF
--- a/ledger-core/virtual/object/object.go
+++ b/ledger-core/virtual/object/object.go
@@ -379,27 +379,12 @@ func (sm *SMObject) migrate(ctx smachine.MigrationContext) smachine.StateUpdate 
 		sm.smFinalizer.Report.ProvidedContent.LatestDirtyState = sm.BuildLatestDirtyState()
 	}
 
-	sdlStateReport := ctx.Share(&sm.smFinalizer.Report, 0)
-	if !ctx.Publish(finalizedstate.BuildReportKey(sm.Reference), sdlStateReport) {
-		ctx.Log().Warn(struct {
-			*log.Msg  `txt:"failed to publish state report"`
-			Reference string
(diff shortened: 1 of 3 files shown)


ledger-core/virtual/object/object.go
```diff
@@ -489,3 +474,33 @@ func (sm *SMObject) checkPendingCounters(logger smachine.Logger) {
 		})
 	}
 }
+
+func (sm *SMObject) sharedAndPublishStateReport(
```
cyraxred, line 478: this method is better to return "error"

ledger-core/virtual/object/object.go
```diff
@@ -489,3 +474,33 @@ func (sm *SMObject) checkPendingCounters(logger smachine.Logger) {
...
+	if !ctx.Publish(finalizedstate.BuildReportKey(sm.Reference), sdlStateReport) {
+		return ctx.Error(throw.New("failed to publish state report", struct {
+			Reference reference.Holder
+		}{sm.Reference}))
+	}
+
+	return smachine.StateUpdate{}
+}
+
+func (sm *SMObject) sharedAndPublishSMCallSummarySyncLink(
```
cyraxred, line 493: same here

ledger-core/virtual/object/object.go
dgovorukhin: Fixed

ledger-core/virtual/object/object.go
dgovorukhin: Fixed

PR #500, opened 2020-07-14
PLAT-651: use map[pulse]counter to speed up calculation of earliest pulse

ruz:
old code was slowing down with increase of number of processed requests


sexyhamster:
probably need to update PendingList too. Just to keep the things in sync.

PR #501, opened 2020-07-14
NOISSUE: Perf test for SlotMachine

cyraxred:
* SlotMachine perf tests
* Minor perf fix

### DIFF
--- a/ledger-core/conveyor/smachine/perf/perf_test.go
+++ b/ledger-core/conveyor/smachine/perf/perf_test.go
@@ -1,0 +1,183 @@
+// Copyright 2020 Insolar Network Ltd.
+// All rights reserved.
+// This material is licensed under the Insolar License version 1.0,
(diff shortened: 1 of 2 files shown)


ledger-core/conveyor/smachine/perf/perf_test.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+func runBenchSlotMachine(scanCountLimit int, cycleCount int, boost bool, initFn func(*smachine.SlotMachine)) {
+	boostDuration := 10 * time.Millisecond
+	if !boost {
+		boostDuration = 0
+	}
+
+	signal := synckit.NewVersionedSignal()
+	m := smachine.NewSlotMachine(smachine.SlotMachineConfig{
```
sexyhamster, line 84: defer m.Stop() :octocat:

cyraxred: not needed here - because there is no worker routine

PR #504, opened 2020-07-14
PLAT-543: check that validated state equals to last dirty satet in st…

ivanshibitov:
Check that validated state equals to last dirty state in state report.
Skipped up to completing of https://insolar.atlassian.net/browse/PLAT-402


ledger-core/virtual/integration/statereport_test.go
```diff
@@ -179,3 +183,69 @@ func TestVirtual_VStateReport_BadState_StateAlreadyExists(t *testing.T) {
...
+	typedChecker := server.PublisherMock.SetTypedChecker(ctx, mc, server)
+	typedChecker.VStateReport.Set(func(report *payload.VStateReport) bool {
+		require.Equal(t, report.ProvidedContent.LatestDirtyState, content.LatestDirtyState)
+		require.Equal(t, report.ProvidedContent.LatestValidatedState, content.LatestDirtyState)
+		return true
+	})
+
+	wait := server.Journal.WaitStopOf(&handlers.SMVStateReport{}, 1)
+	server.IncrementPulseAndWaitIdle(ctx)
+	waitWithTimeout(wait)
```
sexyhamster, line 239: whats wrong with testutils.WaitSignalsTimed?

ledger-core/virtual/integration/statereport_test.go
```diff
@@ -179,3 +183,69 @@ func TestVirtual_VStateReport_BadState_StateAlreadyExists(t *testing.T) {
...
+		}
+
+		wait := server.Journal.WaitStopOf(&handlers.SMVStateReport{}, 1)
+		server.SendPayload(ctx, payload)
+		waitWithTimeout(wait)
+	}
+
+	typedChecker := server.PublisherMock.SetTypedChecker(ctx, mc, server)
+	typedChecker.VStateReport.Set(func(report *payload.VStateReport) bool {
+		require.Equal(t, report.ProvidedContent.LatestDirtyState, content.LatestDirtyState)
```
sexyhamster, line 232: you can actually not only make sure, that report equals required. You can set chan here to await after pulse change.
And you don't need resending StateReport, so please `return false`

ledger-core/virtual/integration/statereport_test.go
ivanshibitov: yes, I forget about it

ledger-core/virtual/integration/statereport_test.go
ivanshibitov: no problem

PR #506, opened 2020-07-14
PLAT-655: fix TestDelegationToken_SuccessCheckCorrectToken

MikhailZhukov:
Fix TestDelegationToken_SuccessCheckCorrectToken


ledger-core/virtual/integration/delegationtoken_test.go
```diff
@@ -64,7 +64,9 @@ func TestDelegationToken_SuccessCheckCorrectToken(t *testing.T) {
 		t.Run(testMsg.name, func(t *testing.T) {
 			mc := minimock.NewController(t)
 
-			server, ctx := utils.NewUninitializedServer(nil, t)
+			server, ctx := utils.NewUninitializedServerWithErrorFilter(nil, t, func(s string) bool {
+				return false
```
ElSacramento, line 68: can we filter specific error(s)? not all

sexyhamster: We need to retrospect all the possible errors that are happening in all SMs. Which is not very convenient :( In this test we are making sure that we do not get DT handling error. All the others are outside of the scope. We can actually check with strings.Contains for specific error

PR #507, opened 2020-07-14
PLAT-376: ignore response error

sexyhamster:
Ignore error about not sending response, because it is outside of test scope. Response is sent after wg is already done. So it can try to send to already closed network handler.


ledger-core/network/hostnetwork/flaky_test.go
```diff
@@ -22,11 +23,11 @@ import (
 )
 
 func TestHostNetwork_SendRequestPacket2(t *testing.T) {
-	t.SkipNow() // TODO: PLAT-376 this test or network/pool should be fixed
-
 	defer testutils.LeakTester(t)
-	instestlogger.SetTestOutput(t)
-
+	// response could be sent when test is already think that it is finished
```
afiskon, line 27: Something wrong with the grammar. What you meant probably was "when test already thinks that it's finished"

PR #513, opened 2020-07-15
NOISSUE: remove load

skudasov:
remove load


afiskon:
@skudasov where it's located now? I believe you forgot to modify .github/workflows/ve-performance-checks.yaml

PR #522, opened 2020-07-15
PLAT-375: test no dead lock on semaphore

ivanshibitov:
1. Add test on simple dead lock detection on parent semaphore
2. tests skipped since I must be changed after PLAT-432: parent semaphore must be set to 1


ledger-core/virtual/integration/contract_calls_test.go
```diff
@@ -77,6 +77,159 @@ func assertVCallRequest(t *testing.T,
...
+				})
+
+				typedChecker.VCallResult.Set(func(res *payload.VCallResult) bool {
+					switch res.Callee {
+					case objectAGlobal:
+						require.Equal(t, []byte("finish A.Foo"), res.ReturnArguments)
+						assertVCallResult(t, res, server.GlobalCaller(), objectAGlobal, test.flagsA, outgoingA)
+					case objectBGlobal:
+						require.Equal(t, []byte("finish B.Bar"), res.ReturnArguments)
+						assertVCallResult(t, res, objectAGlobal, objectBGlobal, test.flagsA, outgoingA)
```
ElSacramento, line 202: flagsB ?

ledger-core/virtual/integration/contract_calls_test.go
```diff
@@ -77,6 +77,159 @@ func assertVCallRequest(t *testing.T,
 
 }
 
+func Test_NoDeadLock_WhenOutgoingComeToSameNode(t *testing.T) {
+	t.Skip("https://insolar.atlassian.net/browse/PLAT-432")
```
ruz, line 81: no t.Log with link to test case in TR.

ruz: I see cases logged below, this skip should be placed after logging test case.

ivanshibitov: done

PR #523, opened 2020-07-15
NOISSUE: old code was resulting in never stopping SM if it came to present pulse

ruz:
default migration is set to "jump(process)" and then never changed to
other

new code:
* default migration is Stop
* apply custom `jump(process)` migration only to stepWait using JumpExt


aalekseev-ru:
it would be grate to have test for this situation

PR #526, opened 2020-07-16
PLAT-402: Pass LatestValidatedState in VStateReport

dgovorukhin:
https://insolar.atlassian.net/browse/PLAT-402


ledger-core/virtual/object/finalizedstate/statefinalizer.go
```diff
@@ -69,6 +69,7 @@ func (sm *SMStateFinalizer) stepSendVStateReport(ctx smachine.ExecutionContext)
 
 	msg := sm.Report
 
+	msg.ProvidedContent.LatestValidatedState = msg.ProvidedContent.LatestDirtyState
```
ita-sammann, line 72: I think that it is a bad place to set ValidatedState.
Maybe better do it in SMObject (see `BuildLatestDirtyState()` and `BuildStateReport()` for reference)?\

I mean that in future LatestValidatedState will differ from LatestDirtyState.

dgovorukhin: Fixed

PR #529, opened 2020-07-16
NOISSUE: Improve prev pulse handling

cyraxred:
* Added method to properly get prev _operational_ pulse
* Corrected test server to avoid "hanging/first" pulse issue


ledger-core/virtual/integration/utils/server.go
```diff
@@ -152,6 +152,7 @@ func newServerExt(ctx context.Context, t Tester, errorFilterFn logcommon.ErrorFi
 	s.pulseManager = PulseManager
 	s.pulseStorage = Pulses
 	s.pulseGenerator = testutils.NewPulseGenerator(10)
+	s.incrementPulse(ctx) // for sake of simplicity make sure that there is no "hanging" first pulse
```
bigbes, line 155: Can we move that part to init?

sexyhamster: this one could be removed, because in Init we already do IncerementPulse

cyraxred: we need 2 increments (first will have no valid prev)

cyraxred: moved to init

PR #545, opened 2020-07-17
NOISSUE: add `prefix_tree_slow_test.go` tests to longtests   

ita-sammann:
Also, fix Mergeable rules so that assignee's review is required.

### DIFF
--- a/ledger-core/ledger/jet/prefix_tree_slow_test.go
+++ b/ledger-core/ledger/jet/prefix_tree_slow_test.go
@@ -3,7 +3,7 @@
 // This material is licensed under the Insolar License version 1.0,
 // available at https://github.com/insolar/assured-ledger/blob/master/LICENSE.md.
 
-// +build slowtest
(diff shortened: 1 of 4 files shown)


ledger-core/ledger/jet/prefix_tree_slow_test.go
```diff
@@ -3,7 +3,7 @@
 // This material is licensed under the Insolar License version 1.0,
 // available at https://github.com/insolar/assured-ledger/blob/master/LICENSE.md.
 
-// +build slowtest
+// +build longtest
```
cyraxred, line 6: long tests are to be executed multiple times
this test is just veeeery slow, but it doesn't need to be repeated
so should be either
// +build slowtest,!longtest 
or something else

ita-sammann: Ok, I'll make another tag and will add another test run to this workflow

PR #555, opened 2020-07-20
PLAT-432: release semaphore for outgoing

aalekseev-ru:
Release global semaphore while waiting outgoing result.
Reacquire it when getting result.


ledger-core/virtual/execute/execute.go
```diff
@@ -65,6 +65,7 @@ type SMExecute struct {
 	messageSender         messageSenderAdapter.MessageSender
 	pulseSlot             *conveyor.PulseSlot
 	authenticationService authentication.Service
+	globalSem             conveyor.ParallelProcessingLimiter
```
ita-sammann, line 68: Should it be `RunnerLimiter`?

ledger-core/virtual/execute/execute.go
```diff
@@ -65,6 +66,7 @@ type SMExecute struct {
 	messageSender         messageSenderAdapter.MessageSender
 	pulseSlot             *conveyor.PulseSlot
 	authenticationService authentication.Service
+	globalSem             tool.RunnerLimiter
```
bigbes, line 69: please, rename it to `globalSemaphore`

ledger-core/virtual/execute/execute.go
```diff
@@ -67,6 +68,7 @@ type SMExecute struct {
 	messageSender         messageSenderAdapter.MessageSender
 	pulseSlot             *conveyor.PulseSlot
 	authenticationService authentication.Service
+	globalSemaphore       tool.RunnerLimiter
```
sexyhamster, line 71: maybe call it executionLimiter or something like that? semaphore with type that is not semaphore, when we have semaphore as different type.

aalekseev-ru: see conversation above, renamed it after @bigbes  request

PR #561, opened 2020-07-20
PLAT-667: calling method on empty object leads to an error

bigbes:
Added new conditional (`pendingConstructorFinished`) that can be acquired after `readyToWork`.
Moved `stepReleaseExecutionPaths` into bargeIn.
Added "async unlocking" for both `releaseOrderedExecutionPath`/`releaseUnorderedExecutionPath`


ledger-core/virtual/object/object.go
```diff
@@ -332,42 +333,53 @@ func (sm *SMObject) stepWaitState(ctx smachine.ExecutionContext) smachine.StateU
...
+	state := sm.GetState()
+
 	if sm.PreviousExecutorOrderedPendingCount == 0 {
-		sm.releaseOrderedExecutionPath(ctx)
-		sm.releaseUnorderedExecutionPath(ctx)
-	} else if sm.GetState() != Empty {
-		sm.releaseUnorderedExecutionPath(ctx)
+		sm.releaseOrderedExecutionPath()
+		sm.releaseUnorderedExecutionPath()
+	} else if state != Empty {
```
MikhailZhukov, line 341: PreviousExecutorOrderedPendingCount type is uint8. It is always == 0 || > 0. Maybe we can combine this condition with next one?

ledger-core/virtual/object/object.go
```diff
@@ -332,42 +333,53 @@ func (sm *SMObject) stepWaitState(ctx smachine.ExecutionContext) smachine.StateU
...
-	}
+		var pendingConstructorFinishedCtl smsync.BoolConditionalLink
+		if state == Empty {
+			pendingConstructorFinishedCtl = smsync.NewConditionalBool(false, "pendingConstructorFinished")
+			sm.PendingConstructorFinished = pendingConstructorFinishedCtl.SyncLink()
+		}
 
-	return ctx.Jump(sm.stepReadyToWork)
-}
+		sm.SignalOrderedPendingFinished = ctx.NewBargeInWithParam(func(interface{}) smachine.BargeInCallbackFunc {
```
MikhailZhukov, line 352: Does we must register this bargin if state == Ready?

bigbes: Yes, it would notify (through bargeIn) that pending was finished and will release orderedSemaphore

ledger-core/virtual/object/object.go
bigbes: Well, we can't if I understand it right

bigbes: Oh, I've understand what you mean

bigbes: fixed

ledger-core/virtual/execute/execute.go
```diff
@@ -223,36 +239,42 @@ func (s *SMExecute) stepWaitObjectReady(ctx smachine.ExecutionContext) smachine.
...
+		}))
+		return ctx.Jump(s.stepSendCallResult)
+	case object.Inactive:
+		panic(throw.NotImplemented())
+	case object.HasState:
+		// ok
+	}
 
-	s.execution.ObjectDescriptor = objectDescriptor
+	return ctx.Jump(s.stepWaitPendingConstructorFinished)
```
sexyhamster, line 267: I don't like the idea of always going to stepWaitPendingConstructorFinished, even if object is not pending consturctor. The semantics of this is just wrong. I'd rather have
```
if s.pendingConstructorFinished.IsZero() {
  return ctx.Jump(s.stepWaitPendingConstructorFinished)
}
return ctx.Jump(s.stepIsolationNegotiation)
```

So reading the log will clearly distinguish pending constructors from not pending constructors.

ledger-core/virtual/execute/execute_test.go
```diff
@@ -834,7 +834,7 @@ func TestSendVStateReportWithMissingState_IfConstructorWasInterruptedBeforeRunne
 			return false
 		}
 	})
-	virtualTestutils.WaitSignalsTimed(t, 10*time.Second, vStateReportRecv)
+	virtualTestUtils.WaitSignalsTimed(t, 10*time.Second, vStateReportRecv)
```
sexyhamster, line 837: maybe WaitSignalsTimed could go to common testutils? Does it have something virtual specific?

bigbes: maybe separate PR

ledger-core/virtual/object/object.go
```diff
@@ -341,42 +342,50 @@ func (sm *SMObject) stepWaitState(ctx smachine.ExecutionContext) smachine.StateU
...
+			sm.PendingConstructorFinished = pendingConstructorFinishedCtl.SyncLink()
+		} else {
+			sm.releaseUnorderedExecutionPath()
+		}
 
-	if sm.PreviousExecutorOrderedPendingCount > 0 {
-		sm.SignalPendingsFinished = ctx.NewBargeIn().
-			WithJump(sm.stepReleaseExecutionPaths)
-	}
+		sm.SignalOrderedPendingFinished = ctx.NewBargeInWithParam(func(interface{}) smachine.BargeInCallbackFunc {
```
sexyhamster, line 359: why it is NewBargeInWithParam and nor BargeIn? the param is interface{} and is unused.

bigbes: And only way to set BargeIn to custom function.

ledger-core/virtual/object/object.go
```diff
@@ -341,42 +342,50 @@ func (sm *SMObject) stepWaitState(ctx smachine.ExecutionContext) smachine.StateU
...
+			return func(ctx smachine.BargeInContext) smachine.StateUpdate {
+				if state == Empty {
+					smachine.ApplyAdjustmentAsync(pendingConstructorFinishedCtl.NewValue(true))
+					sm.releaseUnorderedExecutionPath()
```
sexyhamster, line 363: If execution is blocked before tolerance semaphore then we don't need extra blocking of unordered semaphor, as far as I understood

bigbes: `sm.releaseUnorderedExecutionPath()` - only if `state != Empty`. If it's empty then we won't unblock it.

sexyhamster: Disregard my comment. Unordered is blocked until either constructor is finished in current pulse, or pending constructor result is received from previous executor.

ledger-core/virtual/execute/execute.go
bigbes: fixed

PR #564, opened 2020-07-21
PLAT-407: forbid tolerable + validated isolation

sexyhamster:
Forbid tolerable + validated isolation for now.


ledger-core/virtual/integration/method_test.go
```diff
@@ -1534,3 +1535,179 @@ func TestVirtual_Method_ForObjectWithMissingState(t *testing.T) {
...
+					objectRef,
+					execute.NewStateID(pn, []byte("ok case")),
+					classRef,
+					[]byte("ok case"),
+				)
+			},
+			callResult: []byte("ok case"),
+		},
+		{
+			name:         "Method tolerable + dirty cannot be executed",
```
MikhailZhukov, line 1573: misprint dirty->validation?

ledger-core/virtual/integration/method_test.go
```diff
@@ -1534,3 +1535,179 @@ func TestVirtual_Method_ForObjectWithMissingState(t *testing.T) {
...
+						AddStart(nil, &execution.Update{
+							Type:   execution.Done,
+							Result: result,
+						})
+
+				}
+
+				server.SendPayload(ctx, &pl)
+			}
+
```
MikhailZhukov, line 1700: maybe you need to change pulse and check in VStateReport that object state was(not) changed?

ledger-core/virtual/integration/method_test.go
sexyhamster: fixed

ledger-core/virtual/integration/method_test.go
sexyhamster: don't need to. I'm checking that runner was never called.

ledger-core/virtual/execute/execute.go
```diff
@@ -333,6 +336,14 @@ func negotiateIsolation(methodIsolation, callIsolation contract.MethodIsolation)
 	return res, nil
 }
 
+func checkAllowedIsolation(isolation contract.MethodIsolation) {
```
bigbes, line 339: I don't see a single reason why you've extracted that check to separate function, but it's not important.

sexyhamster: inlined the function

PR #568, opened 2020-07-22
PLAT-671: send VDelegatedRequestFinished after retries for correct pending count

aalekseev-ru:
add send, add log in handler for empty latestState case, actualize test

### DIFF
--- a/ledger-core/virtual/execute/execute.go
+++ b/ledger-core/virtual/execute/execute.go
@@ -31,6 +31,7 @@ import (
 	"github.com/insolar/assured-ledger/ledger-core/virtual/callsummary"
 	"github.com/insolar/assured-ledger/ledger-core/virtual/descriptor"
 	"github.com/insolar/assured-ledger/ledger-core/virtual/object"
+	"github.com/insolar/assured-ledger/ledger-core/virtual/tool"
 )
 
 /* -------- Utilities ------------- */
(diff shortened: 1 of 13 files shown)


ledger-core/virtual/execute/execute.go
```diff
@@ -722,6 +722,8 @@ func (s *SMExecute) stepSendOutgoing(ctx smachine.ExecutionContext) smachine.Sta
 		}
 	} else {
 		if s.outgoingSentCounter >= MaxOutgoingSendCount {
+			// TODO when CallSummary will live longer than one pulse it needs to be updated
+			s.sendDelegatedRequestFinished(ctx, nil)
```
bigbes, line 726: I don't like the idea of setting "sendDelegatedRequestFinished" in every place where work can be stopped prematurely :)
I'd prefer to set "ErrorHandler"/"some kind of destructors" somewhere and catch all that situations at once.

aalekseev-ru: We discuss this with @ruz and decided that this is important for pending tables. Or I can't understand your idea in second sentence.

bigbes: I mean that part of code (send pending request finished) must be executed on every exit scenario after we've received VDelegatedCallRequest

bigbes: But that's a big addition, so I guess only if @ruz is OK

aalekseev-ru: Yeah, we need some kind of destructor for this.

PR #570, opened 2020-07-23
PLAT-674: added benchmark for runner and benchamrks for VE calls

sexyhamster:
Added benchmarks:
- runner
- constructor call
- ordered call
- created 1000 wallets and run get/set


ledger-core/virtual/integration/benchs/basic_test.go
```diff
@@ -82,8 +83,107 @@ func BenchmarkVCallRequestGetMethod(b *testing.B) {
...
+
+	content := &payload.VStateReport_ProvidedContentBody{
+		LatestDirtyState: &payload.ObjectState{
+			Reference: reference.Local{},
+			Class:     class,
+			State:     walletMemory,
+		},
+	}
+
+	report := &payload.VStateReport{
```
dgovorukhin, line 109: Could u please provide 'AsOf' filed, it will strict required. Also, increment pulse before sending state report, we should get report from the previous pulse not in current

ledger-core/virtual/integration/benchs/semi_test.go
```diff
@@ -0,0 +1,142 @@
...
+	pl := payload.VCallRequest{
+		CallType:       payload.CTConstructor,
+		CallFlags:      payload.BuildCallFlags(contract.CallTolerable, contract.CallDirty),
+		Caller:         server.GlobalCaller(),
+		Callee:         walletproxy.GetClass(),
+		CallSiteMethod: "New",
+		Arguments:      insolar.MustSerialize([]interface{}{}),
+	}
+
+	for i := 0; i < 1000; i++ {
```
dgovorukhin, line 60: I prefer to use constant and comment for this purpose. Please no magic numbers)

sexyhamster: the test is run on 1000 wallets. I don't use this 1000 anymore. In what scope do you want to see the constant?

dgovorukhin: Comments will be enough here. In my vision, we always should use constant+comment instead of value in code.

sexyhamster: added comment. I don't understand what exactly you want to achieve with constant here.

ledger-core/runner/bench_test.go
```diff
@@ -0,0 +1,62 @@
...
+		Sequence: 0,
+		Object:   object,
+	}
+
+	b.ResetTimer()
+	b.ReportAllocs()
+
+	c := make(chan struct{})
+
+	for i := 0; i < b.N; i++ {
```
bigbes, line 53: You can, also, add cheap benchmark for parallel calls

sexyhamster: runStart spawns goroutine. So I don't see any point of parallel calls. And it is outside of ticket scope.

bigbes: Just move `Runner` preparation and `executionContext` to separate function, also use `b.Run()` to remove duplication of code

sexyhamster: I don't understand. Is this benchmark wrong? What does parallel benchmark will prove/bench?

ledger-core/virtual/integration/benchs/semi_test.go
cyraxred: agreed with @sexyhamster - placing constant here gives nothing but polluting code with unnecessary content

PR #571, opened 2020-07-24
NOISSUE: app compartment gateway refactoring

AndreyBronin:
- Gateway gets latest pulse from consensus  Chronicles
- Authorize and Bootstrap requests don't sent pulse


ledger-core/network/hostnetwork/packet/packet.proto
```diff
@@ -143,7 +141,7 @@ message AuthorizeResponse {
     string Error = 3;
     Permit Permit = 4;
     uint32 DiscoveryCount = 5;
-    rms.PulseProto Pulse = 6;
+//    rms.PulseProto PulseNumber = 6 [(gogoproto.customtype) = "github.com/insolar/assured-ledger/ledger-core/pulse.Number", (gogoproto.nullable) = false];
```
cyraxred, line 144: jfyi - gogoproto.casttype is better then gogoproto.customtype, as it doesn't require the target type to implement serialization methods.

PR #572, opened 2020-07-24
NOISSUE: remove bad build tags from launchnet.sh

sexyhamster:
Build tags for debug and convlogtxt should be set manually. For functests they should be set in Env.


ivanshibitov:
Why do you think they should be set manually for launchnet.sh?

sexyhamster:
because convlogtxt affects perfomance and while main way to start local network is run launchnet.sh, this kind of tags should be set on purpose manually

ivanshibitov:
Why not to set it for per test? Since it's

PR #573, opened 2020-07-24
NOISSUE: fix_ci_runs_on_wrong_commit

aalekseev-ru:
try to fix problem, when GHA running not on the latest commit. It causes here
https://github.com/insolar/assured-ledger/pull/568/checks?check_run_id=901992719
look at commit hashes in "checkout code" job and in left up corner
image

discussion where i take this solution 
https://github.com/actions/checkout/issues/299

I never done this before, so i hope I do it right

### DIFF
--- a/.github/workflows/linux-checks.yaml
+++ b/.github/workflows/linux-checks.yaml
@@ -18,9 +18,14 @@ jobs:
       GOPATH: /home/runner/work/assured-ledger/assured-ledger/go
     steps:
       - name: Checkout code
-        uses: actions/checkout@master
+        uses: actions/checkout@v2
         with:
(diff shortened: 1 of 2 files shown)


bigbes:
Please, set the right name for PR and body of PR (it'll go to commit name and commit body respectively)

ita-sammann:
@aalekseev-ru, just a reminder: We separate words in a branch name with hyphens, not underscores.

aalekseev-ru:
we have a call and decided to discuss this with team, current solution has its own problems. The key to this problems hides in  "head.sha" and "merge.sha" provided by github.  @ita-sammann has few ideas how to solve it with side scrypts.

PR #574, opened 2020-07-27
PLAT-409: fail if intolerable call with validated state and object have no such state

MikhailZhukov:
Intolerable call with validated state panic with 'Not implemented' if object have no such state


ledger-core/virtual/execute/execute.go
```diff
@@ -273,6 +273,19 @@ func (s *SMExecute) stepWaitPendingConstructorFinished(ctx smachine.ExecutionCon
 		ctx.Log().Test(markerPendingConstructorWait{})
 		return ctx.Sleep().ThenRepeat()
 	}
+	// set descriptor then pending constructor is finished
```
ElSacramento, line 276: when finished?

ledger-core/virtual/integration/method_test.go
```diff
@@ -1779,7 +1814,7 @@ func TestVirtual_Method_IntolerableCallChangeState(t *testing.T) {
 		p1        = server.GetPulse().PulseNumber
 		isolation = contract.MethodIsolation{
 			Interference: contract.CallIntolerable,
-			State:        contract.CallValidated,
+			State:        contract.CallDirty,
```
ElSacramento, line 1817: Why don't you add a ValidatedState to the object? According to documentation, we have intolerable+validated, tolerable+dirty calls in R0.

ledger-core/virtual/execute/execute.go
MikhailZhukov: fix

ledger-core/virtual/integration/method_test.go
MikhailZhukov: done

PR #585, opened 2020-07-29
NOISSUE: improve/cleanup node keeper

cyraxred:
* removed separation of updates of pulse data and node lists causes race (but rare)
* removed some of excessive intermediate representations for nodes and node lists (but not all yet)


ledger-core/network/integration/network_test.go
```diff
@@ -79,6 +74,8 @@ func TestNodeLeave(t *testing.T) {
 
 func TestNodeGracefulLeave(t *testing.T) {
 	//	t.Skip("FIXME node GracefulStop")
+
+	// TODO actual implementation does NOT do graceful leave
```
AndreyBronin, line 78: Let skip this test

ledger-core/network/integration/network_test.go
```diff
@@ -43,15 +37,16 @@ func TestJoinerNodeConnect(t *testing.T) {
 }
 
 func TestNodeConnectInvalidVersion(t *testing.T) {
+	t.Skip("protocol version should not be exposed")
```
AndreyBronin, line 40: is it for security purpose?

cyraxred: not really, just a lazy writer :)
full text is "protocol version should not be exposed to app logic level - it has to be handled/mapped at protocol level and covered with unit tests then"

PR #591, opened 2020-07-30
PLAT-697: fix managing of pending counters

ivanshibitov:
1. remove pending counters 
2. use known table instead of 1.
3. update tests

### DIFF
--- a/ledger-core/virtual/callregistry/working_table.go
+++ b/ledger-core/virtual/callregistry/working_table.go
@@ -31,19 +31,8 @@ func NewWorkingTable() WorkingTable {
 }
 
 func (wt WorkingTable) GetList(flag contract.InterferenceFlag) *WorkingList {
-	if wt.checkInterferenceFlag(flag) {
+	if flag > 0 && int(flag) <= len(wt.requests) {
 		return wt.requests[flag-1]
-	}
-	panic(throw.IllegalValue())
-}
-
-func (wt WorkingTable) checkInterferenceFlag(flag contract.InterferenceFlag) bool {
-	return flag > 0 && int(flag) <= len(wt.requests)
-}
-
-func (wt WorkingTable) CountActive(flag contract.InterferenceFlag) int {
-	if wt.checkInterferenceFlag(flag) {
-		return wt.requests[flag-1].CountActive()
 	}
 	panic(throw.IllegalValue())
 }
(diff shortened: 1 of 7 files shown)


ledger-core/virtual/execute/execute.go
```diff
@@ -540,7 +540,7 @@ func (s *SMExecute) stepStartRequestProcessing(ctx smachine.ExecutionContext) sm
 			panic(throw.Impossible())
 		}
 
-		state.IncrementPotentialPendingCounter(s.execution.Isolation)
+		//state.IncrementPotentialPendingCounter(s.execution.Isolation)
```
sexyhamster: Delete please

ledger-core/virtual/callregistry/working_table.go
```diff
@@ -31,12 +31,23 @@ func NewWorkingTable() WorkingTable {
...
 		return wt.requests[flag-1]
 	}
 	panic(throw.IllegalValue())
 }
 
+func (wt WorkingTable) checkInterferenceFlag(flag contract.InterferenceFlag) bool {
+	return flag > 0 && int(flag) <= len(wt.requests)
+}
+
+func (wt WorkingTable) CountActive(flag contract.InterferenceFlag) int {
```
sexyhamster, line 44: I wonder why you can't use table.GetList(flag).CountActive? This is literally the same

ivanshibitov: It's with checks and shorter to use

sexyhamster: The checks are exactly the same, and I prefer to have only one way of using this lists. Please do not create duplicate functions

ivanshibitov: ok, GetList also has checks.
But, it's still shorter

sexyhamster: Let's avoid creatung duplicates when it is not required.

bigbes: overall I agree with @sexyhamster, you save 10 charachters in 2 places of code, but you've added duplicate code with length > 20 ¯\_(ツ)_/¯

ivanshibitov: It's used more than 30 times is code. It's not 10 charachters.

sexyhamster: Half of them are tests for new function.

ivanshibitov: negotiated

PR #595, opened 2020-07-30
PLAT-685: don't use zero value in call types

afiskon:
As Ruslan suggested, let's not use zero values in messages, to distinguish cases when the field was filled with some value and when developer forgot to fill it.


sexyhamster:
omg, and not a single test is broken! :)

dgovorukhin:
@afiskon Fogot insolar/payload/proto_vnvn.r0.proto:219

afiskon:
@dgovorukhin actually I wanted to make multiple small PRs instead of a single and large one, since I was not sure about the reaction of the team on these changes. But if you insist...

PR #596, opened 2020-07-30
PLAT-705 validated state should become deactivated after pulse change

aalekseev-ru:
add test for this scenario
1. Create object
2. Deactivate object
3. Change Pulse - dirty state migrate to validated state, both become deactivated
4. Send request on Dirty state - get error
5. Send request on Validated state - get error


ledger-core/virtual/integration/deactivate_test.go
```diff
@@ -139,3 +148,156 @@ func TestVirtual_DeactivateObject(t *testing.T) {
...
 }
+
+// 1. Create object
+// 2. Deactivate object
+// 3. Change Pulse - dirty state migrate to validated state, both become deactivated
+// 4. Send request on Dirty state - get error
+// 5. Send request on Validated state - get error
+func TestVirtual_CallMethod_On_DeactivatedObject(t *testing.T) {
+	defer testutils.LeakTester(t)
+	insrail.LogCase(t, "C5506")
```
ElSacramento, line 159: This test duplicates a new version of C5134:  https://insolar.testrail.io/index.php?/cases/view/5134   
The new version will be applied in https://insolar.atlassian.net/browse/PLAT-706

PR #601, opened 2020-07-31
PLAT-696: better VStateReport validation

bigbes:
- extracted to separate file to increase readability in the future (it seems as bad decision to have all that validators in one file)
- little refactoring of VStateReport.Validate, now all status-dependent checks are in it's own functions
- moved all checking subfunctions to struct that they validate (eg VStateReport_ProvidedContentBody)
- added couple more checks (more thorougly checking of (Un)OrderedPending, etc)


ledger-core/insolar/payload/validation.go
```diff
@@ -5,175 +5,6 @@
 
 package payload
 
-import (
-	"github.com/insolar/assured-ledger/ledger-core/vanilla/throw"
-)
-
-type Validate interface {
+type Validatable interface {
```
sexyhamster, line 8: go way it would be Valitor

sexyhamster: but this one is nice too

bigbes: Yes, I've though about it, but on other hand `Validator` is something that validates what's provided, but `Validatable` shows that it can be validated. It's dilemma that i've faced.

dgovorukhin:
1. I prefer that method implementation `Validatable` will be in the top of file (validation_vstatereport.go)
2. In original issue, was talking about check pulse bounds. I think we should check `IsTimePulse` for pulse values.

bigbes:
@dgovorukhin fixed

bigbes:
tests were broken

PR #605, opened 2020-08-03
PLAT-681: test object deactivation with change pulse

MikhailZhukov:
Test: change pulse when half of the deactivation procedure is completed

### DIFF
--- a/ledger-core/virtual/integration/deactivate_test.go
+++ b/ledger-core/virtual/integration/deactivate_test.go
@@ -23,7 +23,8 @@ import (
 	"github.com/insolar/assured-ledger/ledger-core/runner/executor/common/foundation"
 	"github.com/insolar/assured-ledger/ledger-core/runner/requestresult"
 	"github.com/insolar/assured-ledger/ledger-core/testutils"
-	commontestutils "github.com/insolar/assured-ledger/ledger-core/testutils"
+	commonTestUtils "github.com/insolar/assured-ledger/ledger-core/testutils"
+	"github.com/insolar/assured-ledger/ledger-core/testutils/synchronization"
 
 	"github.com/insolar/assured-ledger/ledger-core/testutils/gen"
 	"github.com/insolar/assured-ledger/ledger-core/testutils/insrail"
@@ -176,7 +177,7 @@ func TestVirtual_CallMethod_On_CompletelyDeactivatedObject(t *testing.T) {
 
 			for _, callTypeTest := range callTypeTestCases {
 				t.Run(callTypeTest.name, func(t *testing.T) {
-					defer commontestutils.LeakTester(t)
+					defer commonTestUtils.LeakTester(t)
 
 					mc := minimock.NewController(t)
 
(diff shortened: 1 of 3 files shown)


ledger-core/virtual/integration/deactivate_test.go
```diff
@@ -394,10 +391,241 @@ func TestVirtual_CallMethod_On_DeactivatedDirtyState(t *testing.T) {
...
+				require.NotNil(t, report.ProvidedContent)
+				require.NotNil(t, report.ProvidedContent.LatestDirtyState)
+				require.False(t, report.ProvidedContent.LatestDirtyState.Deactivated)
+				require.Equal(t, []byte(origDirtyMem), report.ProvidedContent.LatestDirtyState.State)
+				require.NotNil(t, report.ProvidedContent.LatestValidatedState)
+				require.False(t, report.ProvidedContent.LatestValidatedState.Deactivated)
+				require.Equal(t, []byte(origDirtyMem), report.ProvidedContent.LatestValidatedState.State)
+				return true
+			})
+			typedChecker.VDelegatedCallRequest.SetResend(true)
```
ElSacramento, line 538: Is it necessary to resend DelegatedCall? Can we not test SMDelegatedTokenRequest here?

ledger-core/virtual/integration/deactivate_test.go
```diff
@@ -394,10 +391,241 @@ func TestVirtual_CallMethod_On_DeactivatedDirtyState(t *testing.T) {
...
+			typedChecker.VDelegatedCallResponse.SetResend(true)
+			typedChecker.VDelegatedRequestFinished.Set(func(finished *payload.VDelegatedRequestFinished) bool {
+				require.NotNil(t, finished.LatestState)
+				require.True(t, finished.LatestState.Deactivated)
```
ElSacramento, line 542: check `finished.LatestState.State` please

ledger-core/virtual/integration/deactivate_test.go
MikhailZhukov: Yes, it is better to use automatic processing, because it is difficult to create all the internal tables manually. This is why I resend lots of messages to system.

ledger-core/virtual/integration/deactivate_test.go
MikhailZhukov: done

ledger-core/virtual/integration/deactivate_test.go
```diff
@@ -394,10 +391,242 @@ func TestVirtual_CallMethod_On_DeactivatedDirtyState(t *testing.T) {
...
+					CallOutgoing:        server.BuildRandomOutgoingWithPulse(),
+				}
+				server.SendPayload(ctx, &pl)
+			}
+
+			commonTestUtils.WaitSignalsTimed(t, 10*time.Second, executeDone)
+			commonTestUtils.WaitSignalsTimed(t, 10*time.Second, server.Journal.WaitAllAsyncCallsDone())
+
+			// increment pulse twice for stop SMStateFinalizer
+			server.IncrementPulseAndWaitIdle(ctx)
```
ElSacramento, line 620: Why we can't wait for SMStateFinalizer using WaitSignalsTimed ?

MikhailZhukov: We are already do this:
```
stateReportSend := server.Journal.WaitStopOf(&finalizedstate.SMStateFinalizer{}, 1)
...
// increment pulse twice for stop SMStateFinalizer
server.IncrementPulseAndWaitIdle(ctx)
server.IncrementPulseAndWaitIdle(ctx)

commonTestUtils.WaitSignalsTimed(t, 10*time.Second, stateReportSend)
```
But we need to send pulse twice because first one used for send report and create SMStateFinalizer and next one used for stop SMStateFinalizer.

PR #607, opened 2020-08-04
PLAT-688: implement VDelegatedRequestFinshed validation

bigbes:
subj

### DIFF
--- a/ledger-core/insolar/payload/validation_vdelegatedrequestfinished.go
+++ b/ledger-core/insolar/payload/validation_vdelegatedrequestfinished.go
@@ -8,6 +8,7 @@ package payload
 import (
 	"github.com/insolar/assured-ledger/ledger-core/insolar/contract"
 	"github.com/insolar/assured-ledger/ledger-core/pulse"
+	"github.com/insolar/assured-ledger/ledger-core/reference"
 	"github.com/insolar/assured-ledger/ledger-core/vanilla/throw"
 )
 
@@ -27,6 +28,22 @@ func (m *VDelegatedRequestFinished) isIntolerable() bool {
 	return m.GetCallFlags().GetInterference() == contract.CallIntolerable
 }
 
+func isTimePulseBefore(pn pulse.Number, before pulse.Number) bool {
+	return pn.IsTimePulse() && pn.IsBefore(before)
+}
+
+func isSpecialTimePulseBefore(pn pulse.Number, before pulse.Number) bool {
+	return pn.IsSpecial() || pn.IsTimePulse() && pn.IsBefore(before)
+}
+
+func globalBasePulseBeforeOrEqLocalPulse(global reference.Global) bool {
+	var (
+		basePulse  = global.GetBase().GetPulseNumber()
+		localPulse = global.GetLocal().GetPulseNumber()
+	)
+	return basePulse.IsBeforeOrEq(localPulse)
+}
+
 func (m *VDelegatedRequestFinished) Validate(currentPulse pulse.Number) error {
 	if err := m.validateUnimplemented(); err != nil {
 		return err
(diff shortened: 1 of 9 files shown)


ledger-core/insolar/payload/validation_vdelegatedrequestfinished.go
```diff
@@ -0,0 +1,92 @@
...
+		return throw.New("CallFlags interference should be valid")
+	}
+
+	if !m.Callee.IsSelfScope() {
+		return throw.New("Callee must be valid self scoped Reference")
+	} else if pn := m.Callee.GetLocal().Pulse(); !pn.IsTimePulse() || pn >= currentPulse {
+		return throw.New("Callee must have valid time pulse with pulse lesser than current pulse")
+	}
+
+	if m.CallIncoming.IsEmpty() {
```
sexyhamster, line 58: Please, consider using switch for more then if else logic.

ledger-core/insolar/payload/validation_vdelegatedrequestfinished.go
```diff
@@ -0,0 +1,92 @@
...
+		return throw.New("CallIncoming should be non-empty")
+	} else if pn := m.CallIncoming.GetLocal().Pulse(); !pn.IsTimePulse() || pn >= currentPulse {
```
sexyhamster, line 60: pulse.Number has nice readable helpers:
```
IsBefore(pn Number) bool
IsAfter(pn Number) bool
IsBeforeOrEq(pn Number) bool
IsEqOrAfter(pn Number) bool
```
I think it would be easier to read these conditions with helpers

ledger-core/insolar/payload/validation_vdelegatedrequestfinished.go
bigbes: Switch-case doesn't support short variable declaration

ledger-core/insolar/payload/validation_vdelegatedrequestfinished.go
bigbes: `m.CallIncoming.GetBase().Pulse() > m.CallIncoming.GetLocal().Pulse()`
Do you think that's harder to read than
`m.CallIncoming.GetBase().Pulse().IsAfter(m.CallIncoming.GetLocal().Pulse())`
?

bigbes: `pn >= currentPulse` is easier, probably, than `pn.IsAfterOrEq(currentPulse)`, but that will lead to inconsistency in that two cases.

ledger-core/insolar/payload/validation_vdelegatedrequestfinished.go
sexyhamster: well, from what I see this checks could be separated and look easier to read
```
if m.CallIncoming.IsEmpty() {
   return throw.New("CallIncoming should be non-empty")
}

incomingLocalPulse := m.CallIncoming.GetLocal().Pulse()
incomingBasePulse := m.CallIncoming.GetBase().Pulse()

switch {
  case incomingLocalPulse.IsTimePulse() || incomingLocalPulse >= currentPulse:
     return throw.New("CallIncoming local part should have valid time pulse lesser than current pulse")
  case incomingBasePulse.IsTimePulse() || incomingLocalPulse >= currentPulse:
     return throw.New("CallIncoming base part should have valid time pulse lesser than current pulse")
   case incomingBasePulse > incomingLocalPulse:
     return throw.New("CallIncoming base pulse should be less or equal than local pulse")
}
```

then there will be no problems with short variable declarations.

I'm not pushing, but from my point of view it is much easier to read, then long `if else if` with long conditions

ledger-core/insolar/payload/validation_vdelegatedrequestfinished.go
sexyhamster: if you put those function callse to variables
```
incomingLocalPulse.IsAfterOrEq(currentPulse)

incomingBasePulse.IsAfter(incomingLocalPulse)
```
then this will be read like a sentence

ledger-core/insolar/payload/validation_vdelegatedrequestfinished.go
```diff
@@ -0,0 +1,92 @@
...
+	case CTInboundAPICall, CTOutboundAPICall, CTNotifyCall, CTSAGACall, CTParallelCall, CTScheduleCall:
+		return throw.New("CallType is not implemented")
+	case CTInvalid:
+		fallthrough
+	default:
+		return throw.New("CallType must be valid")
+	}
+
+	callFlags := m.GetCallFlags()
+	if f := callFlags.GetState(); f >= contract.StateFlagCount || f <= contract.StateInvalid {
```
sexyhamster, line 46: I'd move this validation to method on stateFlag, so there would be callFlags.GetState().IsValid()

And I'm not sure why these checks are in one if else if

bigbes: Sure, let's do it that way. It's better.
They're grouped since it's one field.

sexyhamster: hm, maybe then do a method on CallFlags?

bigbes: I thought of three methods IsValid on State, IsValid on Interference and IsValid on CallFlags that'll group both.

ledger-core/insolar/payload/validation_vdelegatedrequestfinished.go
bigbes: fixed

ledger-core/insolar/payload/validation_vdelegatedrequestfinished.go
bigbes: fixed

ledger-core/insolar/payload/validation_vdelegatedrequestfinished.go
bigbes: fixed

PR #609, opened 2020-08-04
PLAT-592: attempt to fix TestHostNetwork_WrongHandler 

AndreyBronin:
## what I did

use transport.WaitFakeListeners() to wait until two fake transport starts in Suite and listeners will be ready before test run

### DIFF
--- a/ledger-core/network/transport/fake.go
+++ b/ledger-core/network/transport/fake.go
@@ -155,7 +155,7 @@ func WaitFakeListeners(count int32, timeout time.Duration) error {
 
 	go func(ctx context.Context) {
 		for ;atomic.LoadInt32(&listenersCount) != count && ctx.Err() == nil; {
-			<-time.After(time.Microsecond*100)
+			time.Sleep(time.Microsecond*100)
 		}
 		c <- ctx.Err()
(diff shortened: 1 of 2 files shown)


ledger-core/network/transport/fake.go
```diff
@@ -140,3 +145,25 @@ func (f *fakeStreamTransport) Dial(ctx context.Context, address string) (io.Read
...
+// WaitFakeListeners wait for streamTransports will be started
+// this is used to prevent race in tests
+func WaitFakeListeners(count int32, timeout time.Duration) error {
+	c := make(chan error)
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	go func(ctx context.Context) {
+		for ;atomic.LoadInt32(&listenersCount) != count && ctx.Err() == nil; {
+			<-time.After(time.Microsecond*100)
```
cyraxred, line 158: not critical for tests, but better to always avoid using time.After without select from multiple channels
because time.Sleep is cheaper - it doesn't trash heap & scheduler

AndreyBronin: fixed

PR #615, opened 2020-08-05
PLAT-706: update happy-path test and fix small bug

ElSacramento:
- make table test for object deactivation
- fix small bug in SMObject buildVStateReport step


ledger-core/virtual/integration/deactivate_test.go
```diff
@@ -37,51 +34,60 @@ import (
...
 		t.Run(test.name, func(t *testing.T) {
-			defer testutils.LeakTester(t)
-
 			mc := minimock.NewController(t)
 
-			server, ctx := utils.NewServer(nil, t)
+			server, ctx := utils.NewUninitializedServer(nil, t)
 			defer server.Stop()
 
+			runnerMock := logicless.NewServiceMock(ctx, t, func(execution execution.Context) string {
```
bigbes, line 58: Let's use outgoing request reference here

PR #620, opened 2020-08-06
NOISSUE: proper VirtualExecutorForObject

sexyhamster:
use sha256 based target calculation instead of xor, to ensure proper objects distribution in network


ledger-core/appctl/affinity/affinity.go
```diff
@@ -102,6 +107,6 @@ func (jc *Coordinator) VirtualExecutorForObject(
 func xorBytes(dest []byte, b []byte) {
```
cyraxred, line 107: pls delete it

PR #633, opened 2020-08-07
PLAT-691: implement VStateRequest validation

bigbes:
subj


ledger-core/insolar/payload/requestedcontent_test.go
```diff
@@ -13,6 +13,8 @@ import (
 
 func TestStateRequestContentFlags(t *testing.T) {
 	t.Run("single flag", func(t *testing.T) {
+		t.Parallel()
```
cyraxred, line 16: parallel can mess up with log output

bigbes: removed

PR #645, opened 2020-08-11
NOISSUE: (Ledger) jet tree service

cyraxred:
* a service to maintain local jet tree (for 1 LMN configuration) - see treesvc.Service
* this service is available directly to SMs for simplicity, but is coordinated with pulse changes via component management of conveyor


cyraxred:
Merged after 2 days

PR #652, opened 2020-08-12
PLAT-692: implement VCallRequest validation

bigbes:
subj

### DIFF
--- a/ledger-core/insolar/payload/validation_vcallrequest.go
+++ b/ledger-core/insolar/payload/validation_vcallrequest.go
@@ -1,0 +1,135 @@
+// Copyright 2020 Insolar Network Ltd.
+// All rights reserved.
+// This material is licensed under the Insolar License version 1.0,
+// available at https://github.com/insolar/assured-ledger/blob/master/LICENSE.md.
+
+package payload
+
+import (
+	"github.com/insolar/assured-ledger/ledger-core/vanilla/throw"
+)
+
+var _ Validatable = &VCallRequest{}
+
+func validCallType(ct CallTypeNew) error {
+	switch ct {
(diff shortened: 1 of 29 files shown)


ledger-core/insolar/payload/validation_vcallrequest.go
```diff
@@ -0,0 +1,135 @@
...
+
+	var (
+		callerPulse = m.Caller.GetLocal().GetPulseNumber()
+	)
+	if !isSpecialOrTimePulseBeforeOrEq(callerPulse, currentPulse) {
+		return throw.New("Caller should have special or valid time pulse before or equal to current pulse")
+	}
+
+	if !m.Callee.IsSelfScope() {
+		return throw.New("")
```
cyraxred, line 99: empty error name?

ledger-core/insolar/payload/validation_vcallrequest.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	var (
+		calleePulse = m.Caller.GetLocal().GetPulseNumber()
+	)
+	if !isSpecialOrTimePulseBeforeOrEq(calleePulse, currentPulse) {
+		return throw.New("")
```
cyraxred, line 106: if this is not an accidental, then it is better to use a dedicated error instance

ledger-core/insolar/payload/validation_vcallrequest.go
bigbes: accident

ledger-core/insolar/payload/validation_vcallrequest.go
bigbes: accident

PR #662, opened 2020-08-13
PLAT-732: fix CallIncoming and CallOutgoing fields in VDelegatedReque…

ivanshibitov:
fix CallIncoming and CallOutgoing fields in VDelegatedRequestFinished - now all references have common type


sexyhamster:
what does this fix fixes?

ivanshibitov:
It fixes nothing. It's just standardisation

PR #663, opened 2020-08-13
PLAT-734: don't save both semaphores in SmExecute. Use only required one

ivanshibitov:
subj


ledger-core/virtual/execute/execute.go
```diff
@@ -45,8 +45,7 @@ type SMExecute struct {
 
 	// internal data
 	pendingConstructorFinished smachine.SyncLink
-	semaphoreOrdered           smachine.SyncLink
-	semaphoreUnordered         smachine.SyncLink
+	semaphoreExecute           smachine.SyncLink
```
sexyhamster, line 48: unused

ledger-core/virtual/execute/execute.go
```diff
@@ -892,8 +889,6 @@ func (s *SMExecute) stepAwaitSMCallSummary(ctx smachine.ExecutionContext) smachi
 		}).TryUse(ctx).GetDecision() {
 		case smachine.NotPassed:
 			return ctx.WaitShared(syncAccessor.SharedDataLink).ThenRepeat()
-		case smachine.Impossible:
```
bigbes, line 895: Why you've done that?

ivanshibitov: Written here: https://insolario.slack.com/archives/CQQ0FSRNW/p1597317405311600

PR #674, opened 2020-08-14
NOISSUE: LRU/FIFO cache with low GC impact

cyraxred:
* Strategy-controlled dual LRU/FIFO cache with low GC (no per-value refs)
* Simple implementation template to mitigate Go's lack of generics


ledger-core/vanilla/cachekit/cache_example.go
```diff
@@ -0,0 +1,150 @@
...
+// Access to the key is always updated.
+func (p *UintCache) Replace(key Key, value Value) bool {
+	idx, ok := p.keys[key]
+	if ok {
+		p.core.Touch(idx)
+		p.putEntry(idx, uintEntry{key, value})
+		return false
+	}
+	idx, _ = p.core.Add()
+	p.keys[key] = idx
```
ruz, line 81: looks like here should be `p.putEntry(idx, uintEntry{key, value})`

ledger-core/vanilla/cachekit/cache_example.go
```diff
@@ -0,0 +1,150 @@
...
+
+// Delete removes key and zero out relevant value. Returns (false) when key wasn't present.
+// Access to the key is not updated. Cache entry will become unavailable, but will only be freed after relevant expiry / eviction.
+func (p *UintCache) Delete(key Key) bool {
+	idx, ok := p.keys[key]
+	if !ok {
+		return false
+	}
+	ce := p.getEntry(idx)
+	delete(p.keys, ce.key)
```
ruz, line 123: what if `ce.key` is not equal to `key`. Assert check?

ledger-core/vanilla/cachekit/cache_example.go
```diff
@@ -0,0 +1,150 @@
...
+	*ce = uintEntry{}
+	return true
+}
+
+func (p *UintCache) putEntry(idx Index, entry uintEntry) {
+	pgSize := cap(p.values[0])
+	pgN := idx / pgSize
+
+	if pgN == len(p.values) {
```
ruz, line 132: assert if `pgN < len()`?

ledger-core/vanilla/cachekit/cache_example.go
```diff
@@ -0,0 +1,150 @@
...
+	UintCache is an example/template implementation of a cache that uses the Core.
+
+	To create a new cache type - copy this file, rename struct, set Key and Value type aliases accordingly, then
+	set zeroValue constant in accordance with Value type.
+
+
+ */
+
+func NewUintCache(cs Strategy) *UintCache {
+	sc := &UintCache{
```
ruz, line 23: `cs/sc` naming is very bad - adds work to reader without any reason. Use `cache`, `res`, or `result` instead of `sc`. `strat` or `strategy` instead of `cs`.

ledger-core/vanilla/cachekit/core.go
```diff
@@ -0,0 +1,376 @@
...
+	// TrimOnEachAddition is called once on Core creation. When result is true, then CanTrimEntries will be invoked on every addition.
+	// When this value is false, cache capacity can be exceeded upto an average size of a generation page.
+	TrimOnEachAddition() bool
+
+	// CurrentAge should provide time marks when this Strategy needs to use time-based retention.
+	CurrentAge() Age
+
+	// GetAllocationPage is called once on Core creation to provide a size of entry pages (# of items).
+	// It is recommended for cache implementation to use same size for paged storage.
+	GetAllocationPage() int
```
ruz, line 27: I like `AllocationPageSize()` name more.

ledger-core/vanilla/cachekit/core.go
```diff
@@ -0,0 +1,376 @@
...
+}
+
+func (p *generation) init(capacity int, loadFence bool) {
+	p.access = make([]uint32, 0, capacity)
+	if loadFence {
+		p.fence = make(map[uint32]struct{}, capacity)
+	}
+}
+
+func (p *generation) Add(index Index, age Age) GenNo {
```
ruz, line 255: public method on a private strucr

bigbes: Public methods on private structures can have two roles:
1. for interfaces
2. to show that methods should be used outside

only problem I see here is inconsistency: `Add` is 'public', but `trimOldest` is not.

ledger-core/vanilla/cachekit/core.go
```diff
@@ -0,0 +1,376 @@
...
+	p.access = p.access[:j]
+	if prev == nil {
+		return
+	}
+
+	switch prevN := len(prev.access); {
+	case prevN == 0:
+		return
+	case prevN < j && prevN + j <= cap(p.access):
+		p.access = append(p.access, prev.access...)
```
ruz, line 369: I don't see `start/end` get updated on merges, looks like bug.

ledger-core/vanilla/cachekit/core.go
ruz: what I meant is that `generation` is tightly coupled with `Core`, can not function outside of it and can not be accessed from outside. There are no public methods that return generation object on the Core API.

cyraxred: I use public here to indicate methods used by another class/struct (Core in this case) as Go has no way to separate private and protected methods

ledger-core/vanilla/cachekit/cache_example.go
cyraxred: imho, when variable exists only for <10 lines then its name should not be misleading, but can have no meaning.

cyraxred: fixed

ledger-core/vanilla/cachekit/cache_example.go
cyraxred: thanks! fixed

ledger-core/vanilla/cachekit/cache_example.go
ruz: I agree in general, in this case I had to decipher `cs/sc` a few times and it was painful...

ledger-core/vanilla/cachekit/cache_example.go
cyraxred: added

ledger-core/vanilla/cachekit/cache_example.go
cyraxred: no real need here - it will fail on the last line

p.values[pgN][idx % pgSize] = entry

ledger-core/vanilla/cachekit/core.go
cyraxred: fixed

ledger-core/vanilla/cachekit/core.go
```diff
@@ -0,0 +1,376 @@
...
+	// GetAllocationPage is called once on Core creation to provide a size of entry pages (# of items).
+	// It is recommended for cache implementation to use same size for paged storage.
+	GetAllocationPage() int
+
+	// NextGenerationCapacity is called on creation of every generation page.
+	// Parameters are length and capacity of the previous generation page, or (-1, -1) for a first page,
+	// It returns a capacity for a new page and a flag when fencing must be applied. A new generation page will be created when capacity is exhausted.
+	// Fence - is a per-generation map to detect and ignore multiple touches for the same entry. It reduced cost to track frequency to once per generations
+	// and is relevant for heavy load scenario only.
+	NextGenerationCapacity(prevLen int, prevCap int) (pageSize int, useFence bool)
```
cyraxred, line 34: split in 2 methods as discussed

ledger-core/vanilla/cachekit/core.go
cyraxred: good catch

cyraxred: there is also event ordering is broken here
fixed

PR #676, opened 2020-08-17
NOISSUE: refactor validation, code unification, more checks

bigbes:
subj

### DIFF
--- a/ledger-core/insolar/payload/validation.go
+++ b/ledger-core/insolar/payload/validation.go
@@ -8,27 +8,87 @@ package payload
 import (
 	"github.com/insolar/assured-ledger/ledger-core/pulse"
 	"github.com/insolar/assured-ledger/ledger-core/reference"
+	"github.com/insolar/assured-ledger/ledger-core/vanilla/throw"
 )
 
 type Validatable interface {
 	Validate(currPulse PulseNumber) error
 }
 
-func isTimePulseBefore(pn pulse.Number, before pulse.Number) bool {
-	return pn.IsTimePulse() && pn.IsBefore(before)
+type PulseValidator interface {
+	ValidateTime(pn pulse.Number, before pulse.Number) bool
+	Relation() string
+}
+
+type ExtendedPulseValidator interface {
+	GetSpecialValidator() PulseValidator
+	ValidateSpecialOrTime(pn pulse.Number, before pulse.Number) bool
+	ValidateTime(pn pulse.Number, before pulse.Number) bool
+	Relation() string
 }
 
 func isSpecialOrTimePulseBefore(pn pulse.Number, before pulse.Number) bool {
 	return pn.IsSpecial() || (pn.IsTimePulse() && pn.IsBefore(before))
 }
 
-func isTimePulseBeforeOrEq(pn pulse.Number, before pulse.Number) bool {
-	return pn.IsTimePulse() && pn.IsBeforeOrEq(before)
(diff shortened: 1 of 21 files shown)


ledger-core/insolar/payload/validation_vstatereport.go
```diff
@@ -6,51 +6,77 @@
...
-	case !object.GetLocal().Pulse().IsTimePulse():
-		fallthrough
-	case object.GetLocal().Pulse() >= currPulse:
-		return throw.New("Object pulse should be valid time pulse less that current pulse")
+	objectPulse, err := validSelfScopedGlobalWithPulseBeforeOrEq(m.Object, currentPulse, "Object")
+	if err != nil {
+		return err
+	}
+
+	validateStatus := map[VStateReport_StateStatus]vStateReportValidateStatusFunc{
```
dgovorukhin, line 31: @bigbes why is it better to use map instead switch/case?

bigbes: Unification of check signatures, lesser code duplication; no other practical reason
switch-case may be still faster, can be rewritten as `func(VStateReport_StateStatus) vStateReportValidateStatusFunc` with switch case for functions (combining that two decisions)

dgovorukhin: What is the duplication code? We had 4 methods and switch/case. now we have an additional type for function + map and the same 4 methods.

bigbes: every `case` have exactly same code with only one difference: callback

ledger-core/insolar/payload/validation.go
```diff
@@ -8,32 +8,200 @@ package payload
...
 }
 
-func isTimePulseBefore(pn pulse.Number, before pulse.Number) bool {
-	return pn.IsTimePulse() && pn.IsBefore(before)
+type PulseValidator interface {
+	ValidateTime(pn pulse.Number, before pulse.Number) bool
+	Relation() string
+}
+
+type ExtendedPulseValidator interface {
```
dgovorukhin, line 23: @bigbes If this interface named "extended" PulseValidator maybe we can inherit it?
Signature methods is matched
```
type ExtendedPulseValidator interface {
	PulseValidator
	GetSpecialValidator() PulseValidator
	ValidateSpecialOrTime(pn pulse.Number, before pulse.Number) bool
}
```

ledger-core/insolar/payload/validation_vstatereport.go
dgovorukhin: This code example the same simple, but no need a new type and map.
```
var err error

	switch m.GetStatus() {
	case Ready:
		err = m.validateStatusReady(objectPulse, currentPulse)
	case Empty:
		err = m.validateStatusEmpty(objectPulse, currentPulse)
	case Missing, Inactive:
		err = m.validateStatusMissingOrInactive()
	default:
		return throw.New("Unexpected state received")
	}

	if err != nil {
		return err
	}
```

ledger-core/insolar/payload/validation.go
bigbes: fixed

ledger-core/insolar/payload/validation_vstatereport.go
bigbes: fixed

PR #677, opened 2020-08-18
PLAT-365: make ExecutionContinue more sync

bigbes:
- also rename Err -> Error
- closes PLAT-654

### DIFF
--- a/ledger-core/runner/requestresult/outgoingresult.go
+++ b/ledger-core/runner/requestresult/outgoingresult.go
@@ -5,7 +5,22 @@
 
 package requestresult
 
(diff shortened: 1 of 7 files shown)


ledger-core/runner/requestresult/outgoingresult.go
```diff
@@ -5,7 +5,22 @@
 
 package requestresult
 
+var EmptyOutgoingExecutionResult = OutgoingExecutionResult{}
```
sexyhamster, line 8: what is it for?

bigbes: Singleton to identify uninitialized result

PR #697, opened 2020-08-24
NOISSUE: benchmarks

sexyhamster:
- added more benchmark scenarios for echo and brief echo
- fixed limitations for MaxIdleConnections per host exceeding ephemeral ports
- added gops by default for debug purposes
- added benchmark run for master checks, so if they would be broken we will get notification.


ledger-core/scripts/insolard/launchnet.sh
```diff
@@ -12,7 +12,7 @@ INSOLAR_LOG_FORMATTER=${INSOLAR_LOG_FORMATTER:-"json"}
 INSOLAR_LOG_LEVEL=${INSOLAR_LOG_LEVEL:-"debug"}
 # we can skip build binaries (by default in CI environment they skips)
 SKIP_BUILD=${SKIP_BUILD:-${CI_ENV}}
-BUILD_TAGS=${BUILD_TAGS:-''}
+BUILD_TAGS=${BUILD_TAGS:-'debug'}
```
aalekseev-ru, line 15: is it OK for everyone? i remember we have some discussion about default options for launchnet, @ivanshibitov i suppose you was involved in it?

aalekseev-ru (approved):

ledger-core/scripts/insolard/launchnet.sh
sexyhamster: before it has convlogtxt, which I didn't like. debug is nice to have and has close to 0 performance effect

ita-sammann (approved):

PR #706, opened 2020-08-26
PLAT-752: handlers for cached memory request

MariaVasilenko:
add handlers, need for typedchecker in test


ledger-core/virtual/integration/memory_request_test.go
```diff
@@ -0,0 +1,263 @@
...
+					Interference: pl.CallFlags.GetInterference(),
+					State:        pl.CallFlags.GetState(),
+				}, nil)
+
+				server.SendPayload(ctx, pl)
+				expectedState = methodState
+			}
+
+			// pending
+			if cases.pending {
```
bigbes, line 121: If for pending test is way too big, I prefer to have separate test for it.

MariaVasilenko: will be done later

bigbes (approved):

ruz (approved):

PR #708, opened 2020-08-27
PLAT-778: run func tests on insolard in cloud mode

ivanshibitov:
1. make test_func_cloud
2. Fix  MakeErrorSerializable: now it manages error correctly


ledger-core/cmd/insolard/commands.go
```diff
@@ -94,7 +94,8 @@ func testCloudCommand() *cobra.Command {
 It is used for development and debugging of smart contracts`,
 		Run: func(cmd *cobra.Command, args []string) {
 			// todo: implement mode
-			global.Fatalm(throw.NotImplemented())
+			//global.Fatalm(throw.NotImplemented())
```
ruz, line 97: please delete two above lines

ledger-core/cmd/insolard/commands.go
```diff
@@ -94,7 +94,7 @@ func testCloudCommand() *cobra.Command {
 It is used for development and debugging of smart contracts`,
 		Run: func(cmd *cobra.Command, args []string) {
 			// todo: implement mode
```
ruz, line 96: this comment should go away

ruz (approved):

ledger-core/application/testutils/launchnet/launchnet.go
```diff
@@ -370,7 +370,13 @@ func startNet() (*exec.Cmd, error) {
 		_ = os.Chdir(cwd)
 	}()
 
-	cmd := exec.Command("./scripts/insolard/launchnet.sh", "-pwdg")
+	args := "-pwdg"
+	cloudMode := os.Getenv("CLOUD_MODE")
+	if strings.ToLower(cloudMode) == "true" {
+		args += "m"
```
sexyhamster, line 376: m?

ledger-core/scripts/insolard/launchnet.sh
```diff
@@ -457,8 +461,17 @@ fi
 trap 'handle_sigchld' SIGCHLD
 
 echo "start discovery nodes ..."
-for i in `seq 1 $NUM_DISCOVERY_NODES`
-do
+if [ "$cloud_mode" == true ]
+then
+  echo "run in cloud mode ( one process )"
+  go run -mod=vendor scripts/generate_insolar_configs.go --base-cloud --num-virtual-nodes=5
```
sexyhamster, line 467: why is 5 hardcoded here?

ledger-core/server/server.go
```diff
@@ -24,6 +24,10 @@ func NewVirtualServer(cfgPath string) Server {
 	return insapp.New(cfgPath, virtual.AppFactory)
 }
 
+func NewVirtualMultiServer(cfgPath string, multiFn insapp.MultiNodeConfigFunc) Server {
```
sexyhamster, line 27: this should no be Virtual. It reads configs and starts N nodes. There could be LMN there.

ledger-core/scripts/generate_insolar_configs.go
```diff
@@ -174,9 +178,45 @@ func prepareBootstrapVars() commonConfigVars {
...
+
+	err = createFileWithDir(fileName, string(rawData))
+	check("failed to create base cloud config: "+fileName, err)
+}
+
 func main() {
 	parseInputParams()
 
+	if baseCloud {
+		generateBaseCloudConfig()
```
sexyhamster, line 216: I'd move this to subcommand

ledger-core/scripts/generate_insolar_configs.go
```diff
@@ -174,9 +178,45 @@ func prepareBootstrapVars() commonConfigVars {
...
+		return
+	}
+
 	validateNodesConfig()
```
sexyhamster, line 220: probably, validateNodesConfig should be done in any case

ledger-core/application/testutils/launchnet/launchnet.go
sexyhamster: if it is cloud mode - then "c" would probably suit better

ivanshibitov: As for me - 'c' - for configs. That is why I used 'm'

ledger-core/scripts/insolard/launchnet.sh
ivanshibitov: foget

ledger-core/scripts/generate_insolar_configs.go
ivanshibitov: what for? It's also insolar node? Why one more file?

sexyhamster: because it calls different function and creates different config. So you don't need any other options, except for baseCloud. You can actually even skip numVirtual, and just glob search directory with configs.

ivanshibitov: 1. I don't see the problem, when one binary generates different configs with different options
2. I like the idea with explicit number of virtual more.
If it's principal - I can change. But I like current solution more

sexyhamster: 1. I don't like the design of function that does 
```
func myFunc(condition bool) {
  if condition {
    return myOtherFunc()
  }
}

func myOtherFunc() {
}
```

It looks bad.

2. then you should not only check virtual, but check all kind of nodes params and read there configs, not only virtual

ivanshibitov: 1. I don't agree
2. Yes, here's ok. I will read all directory

ivanshibitov: fixed
+ move to separate file

ruz (approved):

sexyhamster (approved):

PR #715, opened 2020-08-28
PLAT-783: refactoring of foundation

ivanshibitov:
1. shuffle packages:  
    1. isolation flags goes to separate package 
    2. proxy helper goes to insolar/contract 
    3. machine types go to separate package
2. remove global var currentProxyCtx
3. update proxies and wrappers 
    1. Wrappers takes ProxyHelper as a parameter
    2. Generated object, that stores proxy type, now also contains ProxyHelper( it's set by GetObject method )
4. Func test in cloud mode word


ledger-core/application/cmd/insgocc/builder.go
```diff
@@ -14,13 +14,13 @@ import (
 	"strings"
 
 	"github.com/insolar/assured-ledger/ledger-core/insolar/defaults"
+	_type "github.com/insolar/assured-ledger/ledger-core/runner/machine/type"
```
bigbes, line 17: lets rename it to machinetype

bigbes:
Please, rename `_type` to `machinetype` everywhere. Otherwise - lgtm

bigbes:
Also i've some doubts about `isolation`, but nothing specifiс. If everyone else is ok with it - i'm ok too.

sexyhamster (approved):

bigbes (approved):

PR #717, opened 2020-08-31
PLAT-755: stab structures for collecting transcripts

ruz:
stab structures for collecting transcripts


ElSacramento (approved):

MariaVasilenko (approved):

aalekseev-ru (approved):

ledger-core/virtual/validation/transcript.go
```diff
@@ -0,0 +1,15 @@
+package validation
+
+type Transcript struct {
+	Entries []TranscriptEntry
```
cyraxred, line 4: Considering, that Transcript will be big (~1M), I strongly recommend 
* prefer to use by-value / no pointer / no slice types
* make a paged storage from the very beginning (see lineage.lineRecords for a copy-paste implementation)
* use integer indices

PR #729, opened 2020-09-02
PLAT-795: check not nil

ElSacramento:
check if generations are not nil


ledger-core/vanilla/cachekit/core.go
```diff
@@ -163,6 +163,16 @@ func (p *Core) addToGen(index Index, addedEntry, overflow bool) GenNo {
 func (p *Core) useOrAdvanceGen(age Age, added bool) {
 	trimmedEntries, trimmedGens := false, false
 
+	if p.oldest == nil {
```
cyraxred, line 166: well, it will panic with nil anyway ... here and in the other methods

also, I'd suggest to use switch {} for multiple checks - this is both go-style and easier to read

ElSacramento: I want to know if it's nil as soon as possible.

ledger-core/vanilla/cachekit/core.go
```diff
@@ -121,7 +121,7 @@ func (p *Core) Add() (Index, GenNo) {
 // Delete can ONLY be called once per index, otherwise counting will be broken
 func (p *Core) Delete(idx Index) {
 	p.alloc.Dec(idx) // this will remove +1 for oldest tracking
```
bigbes, line 123: please, fix that commentary. they must be in a single block

ElSacramento: fixed

aalekseev-ru (approved):

cyraxred (approved):

PR #730, opened 2020-09-02
PLAT-771: update cache configuration + add test

ElSacramento:
add mutex, add simple test


aalekseev-ru (approved):

ruz (changes requested):
I belive that RWLock is better in this case. Note that Touch requires W lock.

aalekseev-ru (approved):

ruz (approved):

PR #735, opened 2020-09-03
PLAT-781: mnp configuration

ivanshibitov:
prepare keys, certs and app configs in memory


ledger-core/scripts/insolard/bootstrap_template.yaml
```diff
@@ -1,7 +1,7 @@
 members_keys_dir: "{{ .BaseDir }}/configs/"
 discovery_keys_dir: "{{ .BaseDir }}/reusekeys/discovery/"
 not_discovery_keys_dir: "{{ .BaseDir }}/reusekeys/nodes/"
-keys_name_format: "/node_%02d.json"
+keys_name_format: "/node_%d.json"
```
cyraxred, line 4: btw - why?

ledger-core/instrumentation/insapp/server.go
```diff
@@ -15,26 +15,36 @@ import (
...
-	cfgPath string
-	appFn   AppFactoryFunc
-	multiFn MultiNodeConfigFunc
-	extra   []interface{}
+	cfgPath            string
+	appFn              AppFactoryFunc
+	multiFn            MultiNodeConfigFunc
+	extra              []interface{}
+	certManagerFactory CertManagerFactory
+	keyStoreFactory    KeyStoreFactory
```
cyraxred, line 37: please make CertManagerFactory and KeyStoreFactory as a single functions of "ConfigurationProvider" that can return config, key store and cert manager

it will also be more compatible with PLAT-779 approach

ledger-core/instrumentation/insapp/server.go
```diff
@@ -47,6 +57,14 @@ func (s *Server) readConfig(cfgPath string) configuration.Configuration {
...
 }
 
+func (s *Server) SetKeyStoreFactory(factoryFn KeyStoreFactory) {
+	s.keyStoreFactory = factoryFn
+}
+
+func (s *Server) SetCertManagerFactory(factoryFn CertManagerFactory) {
+	s.certManagerFactory = factoryFn
+}
+
```
cyraxred, line 67: There should be no setters for individual parts. 
Necessary factories should be provided at construction first. Then, if you need to override'em please use "impose" approach (see AppCompartment.SetImposer for reference)

cyraxred (changes requested):

ledger-core/scripts/insolard/bootstrap_template.yaml
ivanshibitov: revert

ledger-core/instrumentation/insapp/server.go
ivanshibitov: rewrote with 'ConfigurationProvider'

ledger-core/instrumentation/insapp/server.go
ivanshibitov: remove setters

ledger-core/instrumentation/insapp/server.go
cyraxred: then why do you keep'em as variables on the server?

ivanshibitov: now only config provider is stored

cyraxred (approved):
Needs some finishing touch, but is ok now :)

ita-sammann (approved):

PR #751, opened 2020-09-08
PLAT-793: add func tests on cloud

ivanshibitov:
Add two 'make' targets
1. test_func_cloud - run func tests on cloud insolar
2. test_func_cloud_with_consensus -  On cloud insolar with consensus


cyraxred (changes requested):
cloud + consensus should run without an external pulsar

cyraxred (approved):

MikhailZhukov (approved):

PR #758, opened 2020-09-09
NOISSUE: enforce Any marshaling

ElSacramento:
check length before unmarshal

### DIFF
--- a/ledger-core/rms/any.go
+++ b/ledger-core/rms/any.go
@@ -55,7 +55,7 @@ func (p *Any) MarshalTo(b []byte) (int, error) {
 	if p.value != nil {
 		return p.value.MarshalTo(b)
 	}
-	_, _, err := UnmarshalType(b, func(u uint64) reflect.Type { return nil })
+	_, _, err := UnmarshalType(b, func(u uint64) reflect.Type { return reflect.TypeOf(struct{}{}) })
 	if err != nil {
 		return 0, err
 	}
@@ -66,7 +66,7 @@ func (p *Any) MarshalToSizedBuffer(b []byte) (int, error) {
 	if p.value != nil {
 		return p.value.MarshalToSizedBuffer(b)
 	}
-	_, _, err := UnmarshalType(b, func(u uint64) reflect.Type { return nil })
+	_, _, err := UnmarshalType(b, func(u uint64) reflect.Type { return reflect.TypeOf(struct{}{}) })
 	if err != nil {
 		return 0, err
 	}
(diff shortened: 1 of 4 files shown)


ledger-core/rms/registry.go
```diff
@@ -120,6 +122,9 @@ func (p *TypeRegistry) GetSpecial(id uint64, special string) reflect.Type {
 }
 
 func UnmarshalType(b []byte, typeFn func(uint64) reflect.Type) (uint64, reflect.Type, error) {
+	if len(b) == 0 {
```
cyraxred, line 125: Wrong. Please don't do it

ledger-core/rms/any.go
```diff
@@ -55,13 +55,21 @@ func (p *Any) MarshalTo(b []byte) (int, error) {
 	if p.value != nil {
 		return p.value.MarshalTo(b)
 	}
+	_, _, err := UnmarshalType(b, func(u uint64) reflect.Type { return reflect.TypeOf(struct{}{}) })
+	if err != nil {
+		return 0, err
+	}
 	return 0, nil
```
cyraxred, line 62: why do you need if here? not just 
return 0, err

ledger-core/rms/any.go
```diff
@@ -55,13 +55,21 @@ func (p *Any) MarshalTo(b []byte) (int, error) {
...
 
 func (p *Any) MarshalToSizedBuffer(b []byte) (int, error) {
 	if p.value != nil {
 		return p.value.MarshalToSizedBuffer(b)
 	}
+	_, _, err := UnmarshalType(b, func(u uint64) reflect.Type { return reflect.TypeOf(struct{}{}) })
+	if err != nil {
+		return 0, err
+	}
 	return 0, nil
```
cyraxred, line 73: same as above

cyraxred (changes requested):

ledger-core/rms/any.go
```diff
@@ -55,13 +55,21 @@ func (p *Any) MarshalTo(b []byte) (int, error) {
 	if p.value != nil {
 		return p.value.MarshalTo(b)
 	}
+	_, _, err := UnmarshalType(b, func(u uint64) reflect.Type { return reflect.TypeOf(struct{}{}) })
```
cyraxred, line 58: pls replace a local closure with a global func that returns a pre-defined value

ledger-core/rms/any.go
ElSacramento: fixed

AndreyBronin (approved):

PR #773, opened 2020-09-14
NOISSUE: fix for long running tests

bigbes:
* pulse time is now 1 second instead of 10 second for VE integration tests (that means delay for waiting of VStateReport 0.1ms instead of 1s)
* fixed one test that relies on 1s delay

### DIFF
--- a/ledger-core/testutils/pulsegenerator.go
+++ b/ledger-core/testutils/pulsegenerator.go
@@ -50,6 +50,10 @@ func (g *PulseGenerator) GetPrevBeat() beat.Beat {
 	return g.prev
 }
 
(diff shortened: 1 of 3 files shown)


ruz (approved):
I'm ok with this solution.

bigbes:
```
> time go test ./... -count=300 -timeout=360m -tags="slowtest longtest" -v -json > 1.log
go test ./... -count=300 -timeout=360m -tags="slowtest longtest" -v -json >   6318.43s user 832.77s system 596% cpu 19:59.81 total
```
20 minutes instead of 2 hours

ivanshibitov (approved):

sexyhamster (approved):

PR #788, opened 2020-09-22
PLAT-829: manual pulsar

ivanshibitov:
1. Update pulse generator
2. Add test for generator
4. Don't change pulses for Benchmark_SinglePulseOnCloud_N


ledger-core/application/testutils/launchnet/cloud.go
```diff
@@ -78,8 +81,43 @@ func (cr *CloudRunner) PrepareConfig() {
 	}
 }
 
+func prepareCloudForOneShotRegime(confProvider *server.CloudConfigurationProvider) server.Server {
```
bigbes: Rename it to `prepareCloudForOneShotMode`, please :)

bigbes (dismissed):

ledger-core/application/testutils/launchnet/cloud.go
```diff
@@ -78,8 +82,44 @@ func (cr *CloudRunner) PrepareConfig() {
...
+
+	return s
+}
+
 func (cr CloudRunner) SetupCloud() (func(), error) {
-	s := server.NewMultiServer(cr.ConfProvider)
+	return cr.SetupCloudCustom(false)
+}
+
+func (cr CloudRunner) SetupCloudCustom(pulsarOneShot bool) (func(), error) {
```
sexyhamster, line 116:
```suggestion
type PulsarMode int

var (
   RegularPulsar PulsarMode = iota
   ManualPulsar
)

func (cr CloudRunner) SetupCloudCustom(mode PulsarMode) (func(), error) {
```

I'd rather have this function like this, instead of bool value

ledger-core/server/server.go
```diff
@@ -22,6 +23,24 @@ func NewNode(cfg configuration.Configuration) Server {
 	return insapp.New(cfg, appFactory)
 }
 
+func NewMultiServerWithoutPulsar(configProvider *CloudConfigurationProvider) (Server, pulsar.SelectivePulseDistributor) { // nolint:interfacer
```
sexyhamster, line 26: Its better not to build another contstructor. You can do something like this:
```
func NewMultiServer(configProvider *CloudConfigurationProvider, extraComponents ...interface{}) Server {
	controller := cloud.NewController()
	if configProvider.GetAppConfigs == nil {
		panic("GetAppConfigs cannot be nil")
	}

	multiFn := func(provider insapp.ConfigurationProvider) ([]configuration.Configuration, insapp.NetworkInitFunc) {
		conf := provider.(*CloudConfigurationProvider)
		return conf.GetAppConfigs(), controller.NetworkInitFunc
	}

	return insapp.NewMulti(
		configProvider,
		appFactory,
		multiFn,
		extraComponents...,
	)
}
```

So pulsar would be optional for running MultiServer

ivanshibitov: Why is it better? It's just one more constructor. I don't think this version is more convenient.

ledger-core/application/testutils/launchnet/cloud.go
ivanshibitov: What are other options? We need only two options so far.

ledger-core/server/server.go
sexyhamster: this allow to add more components, and lowers copypasted code

ledger-core/application/testutils/launchnet/cloud.go
sexyhamster: When you'll read code 
```
runner.SetupCloudCustom(RegularPulsar)
```
or 
```
runner.SetupCloudCustom(ManualPulsar)
```

it is easier to understand what is going on, rather then
```
runner.SetupCloudCustom(false)
```

If you want, PulsarMode could be bool, so using named constants will improve code readability

ledger-core/server/server.go
ivanshibitov: 1. Do we need it?
2. I need to return controller or somehow join it one of the extraComponents ( to have opportunity to change pulses )

ledger-core/application/testutils/launchnet/cloud.go
ivanshibitov: But that works only for highest level. On lover levels it will look like runner.SetupCloudCustom(PulsarMode) - it doesn't give you more information.
But ok. I can do it

ledger-core/server/server.go
sexyhamster: 1. yes, it should not be 10 of copypasted constructors
2. you can create controller outside, it will be started with server and could be used outside

ivanshibitov: 1. They are different
2. And at that case I will need to create controller outside every time? Even when I don't need it outside?

ivanshibitov: Controller is in 'internal' dir

sexyhamster: 1. this is poor design. The name is NewMultiServerWithoutPulsar, why does it return controller? 
It could be NewControlledMultiServer() with external controller, that can manage pulses
2. If you start NewMultiServer than yoo don't need controller
3. It could be moved from internal to testutils

ivanshibitov: 1. It doesn't return controller. It returns SelectivePulseDistributor
2. yes, but you sed that controller should be created outside and forwarded to constructor

sexyhamster: 1. I don't understand why. Because controller could be used not only as PulseDistributor (I'm working on it)
2. I'm against NewMultiServerWithoutPulsar, but NewControlledMultiServer with controller as argument looks OK

ivanshibitov: ок

ledger-core/server/server.go
```diff
@@ -22,6 +22,23 @@ func NewNode(cfg configuration.Configuration) Server {
 	return insapp.New(cfg, appFactory)
 }
 
+func NewControlledMultiServer(controller cloud.Controller, configProvider *CloudConfigurationProvider) Server { // nolint:interfacer
```
sexyhamster, line 25: nolint?

ledger-core/testutils/cloud/controller_test.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	controller := cloud.NewController()
+	s := server.NewControlledMultiServer(controller, confProvider)
+	go func() {
+		s.Serve()
+	}()
+
+	// wait for starting all components
+	for !s.(*insapp.Server).Started() {
```
sexyhamster, line 63: this looks ugly. We need to either broaden Server interface or return *insapp.Server so we don't need this

ledger-core/server/server.go
ivanshibitov: Yes, because it fails without that

ledger-core/server/server.go
```diff
@@ -22,6 +22,23 @@ func NewNode(cfg configuration.Configuration) Server {
 	return insapp.New(cfg, appFactory)
 }
 
+func NewControlledMultiServer(controller cloud.Controller, configProvider *CloudConfigurationProvider) *insapp.Server { // nolint:interfacer
```
sexyhamster, line 25: there are no interfaces here, so nolint is redundant.

ledger-core/testutils/pulsegenerator.go
```diff
@@ -6,74 +6,110 @@
...
+func makePacket(delta uint16, b beat.Beat) pulsar.PulsePacket {
+	var entropy rms.Entropy
+	copy(entropy[:], b.PulseEntropy[:rms.EntropySize])
+	return pulsar.PulsePacket{
+		PulseNumber:      b.PulseNumber,
+		Entropy:          entropy,
+		NextPulseNumber:  b.PulseNumber + pulse.Number(delta),
+		PrevPulseNumber:  b.PulseNumber - pulse.Number(delta),
+		EpochPulseNumber: b.PulseNumber.AsEpoch(),
+		PulseTimestamp:   b.StartedAt.UnixNano(),
```
sexyhamster, line 54: UnixNano?

ledger-core/instrumentation/insapp/server.go
```diff
@@ -181,3 +184,7 @@ func (s *Server) Stop() {
 	s.gracefulStop <- syscall.SIGQUIT
 	<-s.waitChannel
 }
+
+func (s *Server) Started() bool {
```
sexyhamster, line 188: I think it's better to have WaitStarted with channel, then have bool function with sleeps

ledger-core/server/server.go
ivanshibitov:
```
golangci-lint run
12
server/server.go:25:60: `configProvider` can be `github.com/insolar/assured-ledger/ledger-core/instrumentation/insapp.ConfigurationProvider` (interfacer)
13
func NewControlledMultiServer(controller cloud.Controller, configProvider *CloudConfigurationProvider) Server {
```

sexyhamster: so why not to have insapp.ConfigurationProvider here?

ivanshibitov: ok. I use interface and that is why removed check

sexyhamster (approved):

cyraxred (approved):
Usability could be improved here, but will do it on actual use

PR #791, opened 2020-09-23
PLAT-824: update tests

MariaVasilenko:


cyraxred:
use of global variables to pass test results is not acceptable

PR #846, opened 2020-10-09
NOISSUE: msgdelivery postponed fix

cyraxred:
* prevent overlapping of buffer
* fix possible lock up of worker on OOB
* improve behavior of multiple workers
* improve retry performance
* better performance of postponed buffers


ledger-core/network/nds/msgdelivery/ctl_retry_msg_worker.go
```diff
@@ -106,78 +121,71 @@ func (p *retryMsgWorker) sendHead(msg *msgShipment, repeatFn func(retries.RetryI
...
 
-func (p *retryMsgWorker) isEmptyPostponed() bool {
-	return p.read == p.write
+func (p *retryMsgWorker) processPostponed(pmsg postponedMsg) {
+	switch {
+	case p.processMsg(pmsg.msg, pmsg.repeatFn):
+	case p.postponedLimit > 1:
+		p.postponedLimit--
+	default:
+		p.postponedLimit = 0
```
dgovorukhin, line 176: Here we can be only if p.postponedLimit already 0, it should be 0, less impossible. Why we need set here?

ledger-core/network/nds/msgdelivery/retries/staged.go
```diff
@@ -23,7 +23,7 @@ type StagedController struct {
 }
 
 type RetryStrategy interface {
-	Retry(ids []RetryID, repeatFn func(RetryID))
+	Retry(ids []RetryID, repeatFn func(RetryID), bulkOverflow func([]RetryID))
```
dgovorukhin, line 26: bulkOverflow is not a clean name for me? What is overflow? "bulkRepeatFn" better.

ledger-core/network/nds/msgdelivery/ctl_retry_msg_worker.go
```diff
@@ -23,26 +23,30 @@ func newRetryMsgWorker(sender *msgSender, parallel, postponed int) *retryMsgWork
...
-	read, write int32
+	// buffer
+	postponed     chan postponedMsg
+	postponedRead <- chan postponedMsg
+	postponedLimit uint32
 
 	wakeup chan struct{}
 }
 
+const minRetryBulkSize = 32
```
dgovorukhin, line 44: What is the heuristic constant? Should it be configurable?

ledger-core/network/nds/msgdelivery/ctl_retry_msg_worker.go
```diff
@@ -106,78 +121,71 @@ func (p *retryMsgWorker) sendHead(msg *msgShipment, repeatFn func(retries.RetryI
...
 		p.marks.unmark(msg.id)
 	default:
 		go p.sendHead(msg, repeatFn)
-		return
+		return true
 	}
 
 	if msg.canSendHead() {
-		p.pushPostponed(msg, repeatFn)
+		postponer.pushPostponed(msg, repeatFn)
```
dgovorukhin, line 147: Why we push message to a worker which busy now? It some balancing mechanism?

dgovorukhin (commented):
I see that nodeMsgMarks copy-pasted and use only pushPostponed method. Maybe we can store function for pushPostponed only?

ledger-core/network/nds/msgdelivery/ctl_retry_msg_worker.go
cyraxred: to get a noticeable difference between n*repeatFn vs bulkFn
low value will cause mostly empty arrays to stuck in stages, high value will increase CPU usage

doesn't need to be configured as almost any value is tolerable

ledger-core/network/nds/msgdelivery/ctl_retry_msg_worker.go
cyraxred: to keep all retries for the same node in a single worker - this will improve responsiveness of other workers on rescanning of postponeds

ledger-core/network/nds/msgdelivery/ctl_retry_msg_worker.go
cyraxred: can get here when postponedLimit == 1

cyraxred: but yes, it can be changed as
	case p.postponedLimit >= 1:
		p.postponedLimit--
    }

ledger-core/network/nds/msgdelivery/retries/staged.go
cyraxred: bulkOverflow is called for entries that were NOT retried, so it is very different than repeatFn 
entries, reported via bulkOverflow will have a bit shorter path to next retry

dgovorukhin: Why interface should know something about self-usage? Overflow on the caller side should not affect the name in the interface. I agree that we need this option, my concern about naming

ledger-core/network/nds/msgdelivery/ctl_retry_msg_worker.go
dgovorukhin: My bad

ledger-core/network/nds/msgdelivery/retries/staged.go
cyraxred: because caller knows about inability of retry (==overflow), while provider knows how to handle it

dgovorukhin (approved):

AndreyBronin (approved):

PR #853, opened 2020-10-12
NOISSUE: rework CachedMemory request/response

ruz:
* Unknown is different from Invalid
* use rms.ObjectState type
* we don't need object reference in response and in request
  we need it only for message routing
* remove redirect, we don't need it at the moment


ElSacramento (approved):

aalekseev-ru (approved):

bigbes (approved):

bigbes (changes requested):
Please, change ID's of message fields to start with 1800 and for records to be more than 40.

PR #854, opened 2020-10-12
NOISSUE: fix too many open files in pulsewatcher

ivanshibitov:
subj


ledger-core/pulsewatcher/runner.go
```diff
@@ -25,15 +25,12 @@ import (
 	"github.com/insolar/assured-ledger/ledger-core/pulsewatcher/status"
 )
 
+var client http.Client
```
sexyhamster, line 28: lets avoid global variable here and just create one client in Run and pass it to CollectNodesStatuses?

ivanshibitov: ok

cyraxred (approved):

PR #878, opened 2020-10-28
PLAT-782: Controllable MNP

sexyhamster:
Controlled MNP runs for jepsen like scenarios.

Cutted corner:
For now concensus run is not supported. It is partially working, but with hacks around pulsar. If we will need them, we'll need to rework single process pulsar to be able to trigger it with ease.

### DIFF
--- a/ledger-core/testutils/cloud/controller.go
+++ b/ledger-core/testutils/cloud/controller.go
@@ -36,43 +36,32 @@ import (
 	"github.com/insolar/assured-ledger/ledger-core/vanilla/throw"
 )
 
-type controlledNode struct {
-	beatAppender               beat.Appender
-	dispatcher                 beat.Dispatcher
-	router                     watermill.Router
-	platformCryptographyScheme cryptography.PlatformCryptographyScheme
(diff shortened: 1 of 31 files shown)


ita-sammann (dismissed):

ledger-core/testutils/cloud/controller.go
```diff
@@ -187,14 +211,11 @@ func (n Controller) PartialDistribute(_ context.Context, packet pulsar.PulsePack
...
 	defer n.lock.Unlock()
 
-	profiles := make([]profiles.StaticProfile, 0, len(n.nodes))
-	for _, netNode := range n.nodes {
-		profiles = append(profiles, netNode.profile)
-	}
+	profiles := n.prepareNodeProfiles()
 
 	// sort profiles by shortID
 	sort.Slice(profiles, func(i, j int) bool {
```
cyraxred, line 221: does it need to be in predictable order?

cyraxred (approved):

ita-sammann (approved):

